PR #421, opened 2023-11-26
fix: use total count to check if there's enough contracts

multikatt:
datasetCount from useContracts() is using the filtered count which makes the onboarding panel appear.


apps/renterd/components/Files/checks/useNotEnoughContracts.tsx
```diff
@@ -3,7 +3,8 @@ import { useRedundancySettings } from '../../../hooks/useRedundancySettings'
 
 export function useNotEnoughContracts() {
   const redundancy = useRedundancySettings()
-  const { datasetCount, isLoading: isContractsLoading } = useContracts()
+  const { dataset, isLoading: isContractsLoading } = useContracts()
```
multikatt, line 6: Im unsure if it's better to it this way or change datasetCount in useContracts() to use the unfiltered dataset, and another variable for pagination called datasetFiltererdCount and update that code as well.

alexfreska: Hey, thanks so much for the PR! I think it would be clearest to have `datasetCount` and `datasetFilteredCount` (and `pageCount`).

multikatt: Alright, I've given that a try as well!

alexfreska (commented):
@multikatt perfect, could you add a changeset so that the bug fix is included in the release notes.
```
npx changeset
```
Use the arrow keys and hit spacebar to select `renterd`, skip major and minor version with enter key to select patch version bump. Enter a message like:
```
Fixed an issue where the filtered contract count was being used in the onboarding checks.
```
And then commit the generated file to this PR, it should be .changesets/some-strange-name.md

alexfreska (approved):
@multikatt looks great! The CI will probably fail because the second two commits are not formatted as conventional commits, maybe squash them into the first commit and force push, after that we are good to merge!

multikatt:
Hope I did that right :smiley:

alexfreska:
Thank you so much! 🚀

PR #452, opened 2024-01-02
fix: website link text

alexfreska:
- Improve links with more verbose text.


n8maninger (approved):
Looks like this got them all 👍🏻

PR #486, opened 2024-02-14
renterd reset lost sector count

alexfreska:
- The host context menu now has an option to reset the lost sector count.

Screenshot 2024-02-14 at 2 03 03 PM

### DIFF
--- a/.changeset/happy-ads-matter.md
+++ b/.changeset/happy-ads-matter.md
@@ -1,0 +1,5 @@
+---
+'renterd': minor
+---
+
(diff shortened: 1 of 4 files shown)


ChrisSchinnerl:
It's registered as an alert. Renterd unregisters alerts upon the next time the check doesn't fail but it won't be instantly.

alexfreska:
@ChrisSchinnerl ok cool, then this change should be sufficient.

ChrisSchinnerl (approved):

PR #487, opened 2024-02-14
add net address and uptime to node profile

alexfreska:
- The renterd and hostd node profile details now include uptime.

Screenshot 2024-02-14 at 2 54 54 PM

### DIFF
--- a/.changeset/gentle-coats-camp.md
+++ b/.changeset/gentle-coats-camp.md
@@ -1,0 +1,5 @@
+---
+'@siafoundation/units': minor
+---
+
+humanTime now supports up to days and long and abbreviated formatting.
--- a/.changeset/hip-ducks-occur.md
+++ b/.changeset/hip-ducks-occur.md
@@ -1,0 +1,5 @@
+---
+'renterd': minor
+---
+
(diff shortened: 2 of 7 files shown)


n8maninger:
Net address doesn’t really make sense for `renterd` imo

ChrisSchinnerl (changes requested):
I agree with Nate. NetAdress doesn't make sense for renters imo.

alexfreska:
@n8maninger @ChrisSchinnerl yeah hmm, I guess that change was mainly for me or others who use the hosted UI with multiple nodes - I was just getting confused as to which node I was looking at. I think `hostd` has a config value for "name", does renterd have this? I could simply show the name, if set.

ChrisSchinnerl:
@alexfreska The bus doesn't have a name but the worker api got a `/id` endpoint that exposes a name that you can give your worker to distinguish between them in a cluster setup.

alexfreska:
@ChrisSchinnerl @n8maninger  I've removed the net address, this is now just uptime.

n8maninger (approved):

PR #488, opened 2024-02-14
hostd integrity check sector data and general dataset refresh intervals

alexfreska:
- Integrity check alert sector data will now update when there is an ongoing integrity check.
- All data will now refresh more frequently.

### DIFF
--- a/apps/hostd/dialogs/AlertsDialog.tsx
+++ b/apps/hostd/dialogs/AlertsDialog.tsx
@@ -12,6 +12,7 @@ import {
 } from '@siafoundation/react-hostd'
 import { humanTime } from '@siafoundation/units'
 import { useCallback } from 'react'
+import { defaultDatasetRefreshInterval } from '../config/swr'
 
 type Props = {
   open: boolean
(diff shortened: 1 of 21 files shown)


n8maninger:
Can everything refresh more frequently? Maybe every 15 seconds?

ChrisSchinnerl (dismissed):

alexfreska:
@n8maninger currently the change is to 20s when viewing alerts, 60s in background. Are you suggesting 15s when viewing?

n8maninger:
Yeah. Im suggesting every data set should refresh at a 10-15s interval.

alexfreska:
@n8maninger alright, updated all apps to use a default of 15s for all primary data.

apps/hostd/dialogs/AlertsDialog.tsx
```diff
@@ -19,7 +20,15 @@ type Props = {
 }
 
 export function AlertsDialog({ open, onOpenChange }: Props) {
-  const alerts = useAlerts()
+  const alerts = useAlerts({
```
n8maninger, line 23: This should always use the default too imo

alexfreska: oop, missed that one, ty.

n8maninger (approved):

PR #496, opened 2024-02-22
feat: renterd multipart upload hooks

alexfreska:
- Add useMultipartUploadCreate, useMultipartUploadComplete, useMultipartUploadAbort, useMultipartUploadAddPart, useMultipartUploadChunk, useMultipartUploadListUploads, useMultipartUploadListParts.

### DIFF
--- a/libs/react-renterd/src/bus.ts
+++ b/libs/react-renterd/src/bus.ts
@@ -29,7 +29,9 @@ import {
   Host,
   HostSettings,
   Obj,
+  PartialSlab,
   SiacoinElement,
+  SlabSlice,
   WalletTransaction,
 } from './siaTypes'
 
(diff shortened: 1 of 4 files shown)


alexfreska:
Current dependencies on/for this PR:
* **#497** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/497?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#496** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/496?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/SiaFoundation/web/496?utm_source=stack-comment).

libs/react-renterd/src/bus.ts
```diff
@@ -838,3 +840,161 @@ export function useMetricsWallet(
 //   hostKey: string
 //   origin: string
 // }
+
+// multipart
+
+export type MultipartUploadCreatePayload = {
+  path: string
+  bucket: string
+  key?: string /// defaults to key:0000000000000000000000000000000000000000000000000000000000000000
```
ChrisSchinnerl, line 849: You don't need to set this anymore. Set `generateKey` to `true` instead to get a random key by default.

libs/react-renterd/src/worker.ts
```diff
@@ -55,6 +55,36 @@ export function useObjectUpload(
...
+export type MultipartUploadChunkParams = {
+  key: string
+  bucket?: string
+  contractset?: string
+  uploadid: string
+  partnumber: number
+  minshards?: number
+  totalshards?: number
+  offset?: number
+  disablepreshardingencryption?: boolean
```
ChrisSchinnerl, line 67: you can drop `disablepreshardingencryption` now too

ChrisSchinnerl (changes requested):

libs/react-renterd/src/worker.ts
alexfreska: Do you mean this option is no longer available? These hooks should mostly match the entire API

libs/react-renterd/src/bus.ts
alexfreska: So I understand, previously I think omitting this key would generate a key, now I pass generateKey: true? But there is still an option to pass a key?

ChrisSchinnerl: Previously omitting the key would use the noop key `key:000000000000000000000000000000000000000000000000000000000000000` which disables the pre-shading encryption.
Since we want that encryption, you need to specify `generateKey: true` for a key to be generated.

libs/react-renterd/src/worker.ts
alexfreska: Cool updating it. Does this mean `offset` is always required then?

ChrisSchinnerl (approved):

PR #509, opened 2024-02-24
fix: renterd storage and upload config without redundancy

alexfreska:
- Storage and upload price settings no longer default to or have the option to include redundancy. The price with the configured redundancy is now always shown below.


alexfreska:
> There seems to be a problem with the estimate in the screenshot? It's rendered twice? I think you rewrote it and copy pasted it twice or something.
> 
> I was thinking maybe it makes sense to mention the word "host" in the description of both the max storage and - upload price. If it would be "The max allowed price a host is allowed to charge to store 1TB worth of data", it would more or less imply that does not include redundancy... because that would not make sense.
> 
> I would have probably made the redundancy line similar to the network average, e.g. `[info-icon] Including Redundancy: 600 SC/TB/Month` and if you hover over the `[info-icon]` you'd get a more thorough explanation of why the configurable price does not include redundancy (so because they're on a per-host basis). Current design is totally fine though, was just a thought.
> 
> Thanks for updating this 👍

@peterjan the repeated estimate was because the first is per TB and the second is for the total estimated storage, when the user has this set to 1TB they look the same, I have updated it to only show if the estimate is not the same.

Your other suggestions also make sense, I have updated the code to match these. The screenshots in the description now reflect these changes.

peterjan:
> @peterjan the repeated estimate was because the first is per TB and the second is for the total estimated storage, when the user has this set to 1TB they look the same, I have updated it to only show if the estimate is not the same.

Cool, maybe it's fine to only show the cost for the total estimated storage. It's pretty easy to work it back to TB/month cost.

peterjan (approved):

ChrisSchinnerl (approved):

PR #519, opened 2024-02-29
refactor: sdk rhp

alexfreska:
- Updated sdk to latest core changes, updated structure.
- Follow up will migrate walletd methods into sdk, use sdk in walletd.


alexfreska:
Current dependencies on/for this PR:
* **#519** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/519?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/SiaFoundation/web/519?utm_source=stack-comment).

n8maninger:
I'm not sure I believe the `rhp` and `wallet` functions deserve their own package. It requires creating small utility packages (`api`, `utils`, `marshal`) to share code.

alexfreska:
> I'm not sure I believe the `rhp` and `wallet` functions deserve their own package. It requires creating small utility packages (`api`, `utils`, `marshal`) to share javascript utilities.
> 
> If we do split them up into distinct packages, I don't think those packages should be "js-aware" and instead have concrete arguments and return values rather than `js.Value` and `api.Result`.

@n8maninger I moved everything into `main` 👍🏼

libs/sdk/src/utils/wasm_exec.js
n8maninger, line 1: This should likely be copied from `$(go env GOROOT)/misc/wasm/wasm_exec.js` when the wasm file is built.

libs/sdk/src/utils/wasm_exec_tinygo.js
n8maninger, line 1: This should probably be deleted, no?

libs/sdk/src/utils/wasm_exec.js
n8maninger: It's also interesting to note that there is a specific `wasm_exec_node.js` for use with node. I'm unsure if we've tested outside of the browser context, but that might be concerning.

ChrisSchinnerl (commented):
I think after addressing Nate's comments this is good to go

libs/sdk/src/utils/wasm_exec.js
alexfreska: https://github.com/SiaFoundation/web/pull/519/files#diff-32383f07904a9dacbd3550fd4d8935af3dfc2211ec49de70a1616dda04be0f1eR18

n8maninger (approved):

ChrisSchinnerl (approved):

PR #537, opened 2024-03-22
fix: handle parameter routes in nextjs router

n8maninger:
Closes https://github.com/SiaFoundation/renterd/issues/1026


alexfreska (approved):
Looks great 🦾

PR #598, opened 2024-04-22
Use path instead of filepath for UI assets

n8maninger:
Switches the nextjs Go router to use `path` instead of `filepath` to prevent issues on Windows. Also adds a few sanity tests to make sure the routers initialize correctly.

### DIFF
--- a/hostd/hostd_test.go
+++ b/hostd/hostd_test.go
@@ -1,0 +1,30 @@
+package hostd
+
+import (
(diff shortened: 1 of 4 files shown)


alexfreska (approved):
👍🏼 I will follow this with updated GitHub actions that run all platform matrix. Also will add changesets to bump the UI modules with this new code.

Issue #596, opened 2024-04-22
File health details show wrong slab keys


ChrisSchinnerl:
As seen in the attached files, all slabs show the key `0419` and partial slabs which should display the same key display different ones. The first screenshot was taken on my own node and the other ones were provided by community member `superbizho` on [Discord](https://discord.com/channels/809849352516141067/946328563320238190/1231911124413448203).

![Image](https://github.com/SiaFoundation/web/assets/3903476/956c17ee-cabd-4455-a9e8-92f30bb324c9)
![Image](https://github.com/SiaFoundation/web/assets/3903476/ee225118-7c32-45a9-87f7-562d6da9ba1e)
![Image](https://github.com/SiaFoundation/web/assets/3903476/b44a83a4-98e8-49be-a731-e12c208a5196)
![Image](https://github.com/SiaFoundation/web/assets/3903476/446b834b-654b-4b46-a405-01607a19aacb)

PR #602, opened 2024-04-22
fix: renterd file health slab keys

alexfreska:
- Fixed an issue where file health slab keys were showing up as repeated and incorrect.

### DIFF
--- a/apps/renterd/components/Files/Columns/FilesHealthColumn/FilesHealthColumnContents.tsx
+++ b/apps/renterd/components/Files/Columns/FilesHealthColumn/FilesHealthColumnContents.tsx
@@ -57,7 +57,12 @@ export function FilesHealthColumnContents({
   const slabs = sortBy(
     obj.data.object.slabs?.map((s) => ({
       ...s.slab,
-      key: `${s.offset}${s.length}${s.slab.key}`,
+      // id is for use as a unique React key.
+      // slab key is not necessarily unique. e.g. an object uploaded with tiny
+      // multipart uploads might reference the same slab over and over but at
+      // different offsets and lengths. So we should not assume that they are
+      // always unique.
+      id: `${s.offset}${s.length}${s.slab.key}`,
       isPartialSlab: !!s.slab.shards,
       contractSetShards: s.slab.shards?.length
         ? computeSlabContractSetShards({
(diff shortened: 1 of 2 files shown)


alexfreska:
* **#602** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/602?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

alexfreska:
@ChrisSchinnerl PR for https://github.com/SiaFoundation/web/issues/596. Just want to to confirm, object slab keys should be unique? As in object will not have 2 slabs with the same key?

apps/renterd/components/Files/Columns/FilesHealthColumn/FilesHealthColumnContents.tsx
```diff
@@ -57,7 +57,6 @@ export function FilesHealthColumnContents({
   const slabs = sortBy(
     obj.data.object.slabs?.map((s) => ({
       ...s.slab,
-      key: `${s.offset}${s.length}${s.slab.key}`,
```
alexfreska, line 60: I think in the past there was a weird issue where some objects had slabs with the same key (so added this "React `key`" for uniqueness, but forgot the name conflicted with the actual `key`), I assume that keys are actually supposed to be unique across the object so removing this temp fix which also fixes the observed issue too.

ChrisSchinnerl:
> @ChrisSchinnerl PR for #596. Just want to to confirm, object slab keys should be unique? As in object will not have 2 slabs with the same key?

It's not impossible but highly unlikely. e.g. an object uploaded with tiny multipart uploads might reference the same slab over and over but at different offsets and lengths.

alexfreska:
> > @ChrisSchinnerl PR for #596. Just want to to confirm, object slab keys should be unique? As in object will not have 2 slabs with the same key?
> 
> It's not impossible but unlikely. e.g. an object uploaded with tiny multipart uploads might reference the same slab over and over but at different offsets and lengths. So we should not assume that they are always unique.

@ChrisSchinnerl ah good to know, I must have hit this issue. I will add back the unique `id` I had but name it `id`.

apps/renterd/components/Files/Columns/FilesHealthColumn/FilesHealthColumnContents.tsx
alexfreska: Nevermind, Chris confirmed that not necessarily unique - adding back.

ChrisSchinnerl (approved):

PR #621, opened 2024-05-07
walletd maturity height, walletd events update

alexfreska:
- Updated wallet events to include maturityHeight and renamed val to data.
- The events list now shows the maturity height, transactions that have not reached this height are shown as locked.
Screenshot 2024-05-08 at 8 54 32 AM

### DIFF
--- a/.changeset/gentle-zoos-love.md
+++ b/.changeset/gentle-zoos-love.md
@@ -1,0 +1,6 @@
+---
+'walletd': minor
+'@siafoundation/walletd-types': minor
+---
+
+Updated wallet events to include maturityHeight and renamed val to data. Closes https://github.com/SiaFoundation/walletd/issues/114
--- a/.changeset/six-fishes-check.md
+++ b/.changeset/six-fishes-check.md
@@ -1,0 +1,5 @@
+---
+'walletd': minor
+---
+
+The events list now shows the maturity height, transactions that have not reached this height are shown as locked. Closes https://github.com/SiaFoundation/walletd/issues/115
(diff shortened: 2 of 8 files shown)


alexfreska:
* **#621** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/621?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

n8maninger:
Does it make sense to show both height and maturity height? That may get confusing to see both when most transactions mature instantly. I think I'd prefer to just show maturity height and label it as height. 

Another question: currently, events are sorted by maturity height. That means immature payouts are always at the top. I'm on the fence on whether that's fine UX or confusing. Maybe they should be sorted by the height they were received instead?

alexfreska:
> Does it make sense to show both height and maturity height? That may get confusing to see both when most transactions mature instantly. My gut says show maturity height labeled as height, but it may make more sense to users the other way around.
> 
> Related: currently, events are sorted by maturity height. That means immature payouts are always at the top. I'm on the fence on whether that's fine UX or confusing. Maybe they should be sorted by the height they were received instead?

@n8maninger
- I can combine the maturity height into one height column that displays both values in some way.
- I think sorting by maturity height makes sense for 90% of users, maybe we could also offer timestamp or height sorting options.

alexfreska:
@n8maninger let me know what you think of the updated screenshot.

n8maninger:
That's super nice. Looks like the amount isn't showing though. The value should be in `data.siacoinOutput.siacoinOutput.value`

alexfreska:
> @alexfreska that's super nice. Looks like the amount isn't showing though. The value should be in `data.siacoinOutput.siacoinOutput.value`

@n8maninger ah good catch, fixed.

n8maninger (approved):

PR #625, opened 2024-05-13
fix: ipv6 connect peer and block host

alexfreska:
> This PR adjusts the client-side validation for IPs.

- The connect to peer dialog now supports IPv6 addresses.
- The host blocklist dialog now supports adding IPv6 addresses.


alexfreska:
* **#625** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/625?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

libs/design-system/src/lib/ipRegex.ts
```diff
@@ -1,7 +1,11 @@
-// Simpler combined regex https://regex101.com/r/0WMysi/2
+const domain = '((?!-)[A-Za-z0-9-]+([-.]{1}[a-z0-9]+)*.[A-Za-z]{2,6})'
```
n8maninger, line 1: Is it worth the hassle to validate this client side? Calling the endpoint will try to connect. If it's not valid, it will fail anyway.

alexfreska: Its a little extra immediate feedback but I agree we could just get rid of it.

What about updating the blocklist, should I remove there too? Does the server-side validate the addresses being added via `/bus/hosts/blocklist`? cc @ChrisSchinnerl

alexfreska: Its extra immediate feedback which I think is nice for the user, but I agree at least in this simple connect form we could get rid of it as the error will be pretty clear.

What about updating the blocklist, should I remove there too? Does the server-side validate the addresses being added via `/bus/hosts/blocklist`? cc @ChrisSchinnerl 

I feel like might as well leave them both, rather than removing, unless you foresee this causing any issues.

ChrisSchinnerl: No, I don't think the blocklist does. It just compares whatever string you provide to whatever string the host announced itself as. e.g. if a host announces as "some.invalid:string" you are able to block that string and prevent that host from being scanned (even though it would fail anyway) even though the address you blocked isn't a valid one.

alexfreska: @ChrisSchinnerl that is good to know, cc @n8maninger - so would you both rather I remove validation in these two places or leave it? Let me know and I will update the PR.

n8maninger: I think removing the validation is fine in both cases

ChrisSchinnerl (commented):
I agree with Nate. Probably fine to drop validation in both places.

alexfreska:
@n8maninger  @ChrisSchinnerl validation removed in both spots, and the regexes deleted forever. Made the connect peer dialog one field instead of separate port.

n8maninger (approved):

PR #644, opened 2024-06-18
refactor: wallet events

alexfreska:
- Wallet events have been updated to use the new types and data format. https://github.com/SiaFoundation/walletd/issues/134


libs/types/src/v2.ts
```diff
@@ -94,4 +96,48 @@ export type AttestationElement = StateElement & {
...
+      renterRollover: Currency
+      hostRollover: Currency
+      renterSignature: Signature
+      hostSignature: Signature
+    }
+  }
+
+export type V2FileContractResolutionDataStorageProof =
+  V2FileContractResolutionTypeBase & {
+    type: 'storage proof'
```
alexfreska, line 129: @n8maninger should this be camel case? wallet event types are camel case.

libs/types/src/v2.ts
```diff
@@ -94,4 +96,48 @@ export type AttestationElement = StateElement & {
...
 
-export type V2FileContractResolutionType = unknown // TODO: replace `unknown` with the actual type
+type V2FileContractResolutionTypeBase = {
+  parent: V2FileContractElement
+}
+
+export type V2FileContractResolutionTypeExpiration =
+  V2FileContractResolutionTypeBase & {
+    type: 'expiration'
+    resolution: Record<string, never> // is this always empty for expiration type?
```
alexfreska, line 106: @n8maninger could you confirm?

libs/types/src/v2.ts
n8maninger: I'd like to be consistent, but these are in `core`. For now we'll leave the inconsistency and discuss in a follow up if that's alright.

libs/types/src/v2.ts
n8maninger: That's correct, the resolution on an "expiration" type is empty

n8maninger (approved):

ChrisSchinnerl (approved):

alexfreska:
### Merge activity

* **Jun 20, 12:40 PM EDT**: @alexfreska started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/SiaFoundation/web/644).

PR #647, opened 2024-06-21
refactor: renterd allowance derived max pricing

alexfreska:
Status: Ready for review

The PR adds a default option to automatically calculate max gouging prices from the user configured usage estimates and allowance.

I don't think the previous simpler `f(allowance, percents) => pricing` version really captured the intended behaviour so I expanded the equation to incorporate the user estimates, while proportionally fixing the prices according to configurable weights, default: 10% upload, 50% download, 40% storage. Let me know if this solution makes sense.

The `derivePricingFromAllowance` method calculates the max price per TB for storage, download, and upload given a total spending allowance and estimated usage in TB for each type. The prices are kept proportional to each other using the weight factors provided as parameters. Storage and upload are also scaled by the redundancy multiplier.

 The total cost equation is:
 ```
allowance * allowanceFactor = scaledAllowance =
   storageTBWithRedundancy * storageWeight * unitCost +
   downloadTBMonth * downloadWeight * unitCost +
   uploadTBMonthWithRedundancy * uploadWeight * unitCost
 ```
 
 Once the unit cost is determined, the individual prices can be calculated:
```
maxUploadPriceTB = unitCost * uploadWeight
maxDownloadPriceTB = unitCost * downloadWeight
maxStoragePriceTBMonth = unitCost * storageWeight
```

The function also includes an allowance scaling factor to account for contract price variation. For example, if a user expects to pay $2 per TB, they should set the max value to $4 per TB because the optimal contracts may vary, eg: one contract at $1 and another at $3 which average to $2. The scaling factor helps avoid unnecessary churn.

Screenshot 2024-07-01 at 5 01 05 PM
Screenshot 2024-07-01 at 5 00 32 PM

### DIFF
--- a/apps/renterd/contexts/config/derivePricesFromAllowance.ts
+++ b/apps/renterd/contexts/config/derivePricesFromAllowance.ts
@@ -1,0 +1,97 @@
+import BigNumber from 'bignumber.js'
+
+/**
+ * This function calculates the max price per TB for storage, download, and
+ * upload given a total spending allowance and estimated usage in TB for each
+ * type. The prices are kept proportional to each other using the weight factors
+ * provided as parameters. Storage and upload are also scaled by the redundancy
+ * multiplier.
+ *
+ * The total cost equation is:
+ * allowance * allowanceFactor = scaledAllowance =
+ *   storageTBWithRedundancy * storageWeight * unitCost +
+ *   downloadTBMonth * downloadWeight * unitCost +
+ *   uploadTBMonthWithRedundancy * uploadWeight * unitCost
+ *
+ * Once the unit cost is determined, the individual prices can be calculated:
+ * maxUploadPriceTB = unitCost * uploadWeight
+ * maxDownloadPriceTB = unitCost * downloadWeight
+ * maxStoragePriceTBMonth = unitCost * storageWeight
+ *
+ * The function also includes an allowance scaling factor to account for
+ * contract price variation. For example, if a user expects to pay $2 per TB,
+ * they should set the max value to $4 per TB because the optimal contracts may
+ * vary, eg: one contract at $1 and another at $3 which average to $2. The
+ * scaling factor helps avoid unnecessary churn.
+ *
+ * @param params - The parameters for the function.
+ * @param params.allowanceMonth - The total spending allowance per month.
+ * @param params.allowanceFactor - The scaling factor to apply to the allowance.
+ * @param params.storageTB - The estimated amount of storage in TB.
+ * @param params.downloadTBMonth - The estimated amount of download in TB per
+ * month.
+ * @param params.uploadTBMonth - The estimated amount of upload in TB per month.
+ * @param params.redundancyMultiplier - The redundancy multiplier.
+ * @param params.storageWeight - The weight factor for storage.
+ * @param params.downloadWeight - The weight factor for download.
+ * @param params.uploadWeight - The weight factor for upload.
+ * @returns An object containing the price per TB for storage, download, and
+ * upload.
+ */
+export function derivePricingFromAllowance({
+  allowanceMonth,
+  allowanceFactor = 1.5,
+  storageTB,
+  downloadTBMonth,
+  uploadTBMonth,
+  redundancyMultiplier,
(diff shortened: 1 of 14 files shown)


alexfreska:
* **#647** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/647?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/renterd/contexts/config/derivePricesFromAllowance.ts
```diff
@@ -0,0 +1,97 @@
...
+    storageTBWithRedundancy
+      .times(storageWeight)
+      .plus(downloadTBMonth.times(downloadWeight))
+      .plus(uploadTBMonthWithRedundancy.times(uploadWeight))
+  )
+
+  // Calculate the price per TB for each type of usage.
+  const maxUploadPriceTB = unitCost.times(uploadWeight)
+  const maxDownloadPriceTB = unitCost.times(downloadWeight)
+  const maxStoragePriceTBMonth = unitCost.times(storageWeight)
```
ChrisSchinnerl, line 90: Dividing by the weights and then multiplying by them again seems a bit less obvious to me than splitting the allowance up first and then dividing by expected usage to get the max values.

e.g.
```typescript
// Calculate dedicated allowance for upload, download and storage
const weightSum = storageWeight + downloadWeight + uploadWeight
const uploadAllowance = scaledAllowance.times(uploadWeight).div(weightSum)
const downloadAllowance = scaledAllowance.times(downloadWeight).div(weightSum)
const storageAllowance = scaledAllowance.times(storageWeight).div(weightSum)

// Figure out max cost that satisfies the corresponding allowance
const maxUploadPriceTB = uploadAllowance.div(uploadTBMonthWithRedundancy)
const maxDownloadPriceTB = downloadAllowance.div(downloadTBMonth)
const maxStoragePriceTBMonth = storageAllowance.div(storageTBWithRedundancy)
```

alexfreska: @ChrisSchinnerl the two approaches are a little different.

Eg: If the user wants to upload and store 1 TB and then download it over and over say 20 TB, the PR version keeps the prices proportional, storage vs download is still close in **_actual price_** (4x vs 5x). The snippet version takes the fixed download budget and drives the price way down relative to the other prices, keeping the **_overall spending_** on each category proportional. I included the example values below. 

Let me know if this makes sense, as I was implementing I switched to the PR version of the equation because it seemed to generate a more intuitive result for the goal, my understanding was that people typically want to price their their upload, storage, download prices relative to each other, but I could be off on this.

Also note that in the snippet version storing 1 TB could potentially cost 80% of the allowance. Leaving only 2k SC for 20TB of download.

```
 {
      allowanceMonth: '10000.00',
      allowanceFactor: 2,
      storageTB: '1.00',
      downloadTBMonth: '20.00',
      uploadTBMonth: '1.00',
      redundancyMultiplier: '1.00',
      storageWeight: 4,
      downloadWeight: 5,
      uploadWeight: 1
    }


snippet version:
maxUploadPriceTB 2000.00 maxDownloadPriceTB 500.00 maxStoragePriceTBMonth 8000.00
totalSpendingOnUpload 2000.00 totalSpendingOnDownload 10000.00 totalSpendingOnStorage 8000.00

PR version:
maxUploadPriceTB 190.48 maxDownloadPriceTB 952.38 maxStoragePriceTBMonth 761.90
totalSpendingOnUpload 190.48 totalSpendingOnDownload 19047.62 totalSpendingOnStorage 761.90
```

alexfreska: yeah, 2 may be be a bit too much, lets try 1.5. I will update that.

ChrisSchinnerl (approved):

alexfreska:
### Merge activity

* **Jul 8, 10:52 AM EDT**: @alexfreska started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/SiaFoundation/web/647).

PR #659, opened 2024-07-10
fix: remove 1 px margin between in scroll bar holder and scroll bar

kocoten1992:
This would fix [issue](https://github.com/SiaFoundation/hostd/issues/423)

The more elegant fix would be:
- create a invisible bar on top the actual scroll bar without the margin.
- and make current scroll bar unclickable, that should keep the visual the same - but would complicated code.

If you like this simple approach then please merge this. Thank you!


alexfreska (approved):
Hi @kocoten1992 thank you for reporting this issue and taking the initiative to open the PR!

I think your change is the correct solution. I don't think there was any reason for the pixel of margin besides how it visually looked, which is not very different from without it - and it certainly causes a UX issue.

Lets merge this. Thank you for the contribution!

PR #671, opened 2024-07-25
feat: renterd price pinning api

alexfreska:
- Add price pinning configuration API.

PR #672, opened 2024-07-25
feat: renterd price pinning

alexfreska:
This PR:
- Adds price pinning support.
- Integrates pinned prices into the automatic price calculation features.
- Adds support for generating recommendations with pinned prices.

https://github.com/user-attachments/assets/e7e3a4ee-2c14-4305-8bd1-1c4a161d1a37

### Pinnable prices
- Allowance, max storage price, max upload price, max download price, and max RPC price fields now support price pinning. The user can choose to pin any or all of those values to a fiat currency value.
- Pinnable inputs display a warning with instructions on how to enable the price pinning feature.
- The configuration has a new section for enabling pinning, selecting the currency, setting the exchange rate endpoint, and the threshold.

### Auto-fit allowance/prices
- This PR refines the "calculate prices" feature and transforms it from a mode that automatically sets read-only price fields into two explicit functions that the user can choose to apply.
- This allows the user to set an allowance they are comfortable with and select "set max prices to fit current allowance". The prices will now fit their spending target. But at this stage, the users spending and pricing configuration may not match with enough hosts, in this situation the system may offer recommendations such as to increase the max download price. The user may decide to apply  this recommendation to match with more hosts. The allowance status would then warn the user that it may not fit their pricing, the user can then use the "fit current max prices" option to bump the allowance up to match the changes. These options allow the user more flexibility and options for tuning in their configuration, guiding them to balance their selections while still matching with enough hosts.

Screenshot 2024-08-02 at 5 27 07 PM
Screenshot 2024-08-02 at 5 30 48 PM

- **Set max prices to fit current allowance**: This action does what "calculate prices" previously did automatically.
  - Set suggested max prices for storage, upload, and download that fit the current allowance spending target. The suggested values take into account the current estimated usage and keeps storage, upload, and download upload prices proportional to each other according to the following weights: 4x storage, 1x upload, 5x download.
- **Set allowance to fit current max prices**: This action does the reverse of the old "calculate prices"
  - Set the allowance to fit the current max prices for storage, upload, and download.
- **Status**: whether the current prices "fit" the allowance.
  -  This is determined calculating the estimated required spending and scaling it by the same allowance factor used for the fit current allowance feature.

Screenshot 2024-08-02 at 6 05 04 PM

- The fit allowance feature also works across a mix of pinned and unpinned prices. The user may opt to pin their storage price but not pin their download price. The fitting will set only the active variant of the field with the equivalent currency value.

Screenshot 2024-08-02 at 5 50 49 PM

- The recommendations now support evaluating pinned fields.
- Each max pricing field that is factored into the allowance now has (when available) both a recommendation for matching with more hosts and an option to fit the current allowance. These are available in both non-pinned siacoin and pinned fiat modes.

### Other
- Add RPC network average tip.
- All tips and recommendations are now displayed according to the app preferences currency setting.


alexfreska:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/672?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#672** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/672?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#671** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/671?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

alexfreska:
> @ChrisSchinnerl check out the preview link, but first this depends on the preview PR change which assumes the `autopilots` key is not null: https://github.com/SiaFoundation/web/pull/671/files#r1709892110

@ChrisSchinnerl I modified the PR to do a null check, so the previews will work now.

ChrisSchinnerl (approved):

PR #685, opened 2024-08-16
fix: renterd pinned field units

alexfreska:
- Fixed an issue where the UI would set extremely small pinned max price values because it set values in per byte instead of TB.
- Max RPC price is no longer pinnable.

### DIFF
--- a/apps/renterd/contexts/config/transformDown.ts
+++ b/apps/renterd/contexts/config/transformDown.ts
@@ -19,8 +19,6 @@ import {
   valuePerBytePerBlockToPerTBPerMonth,
   valuePerPeriodToPerMonth,
   valuePerOneToPerMillion,
-  valuePerByteToPerTB,
-  weeksToBlocks,
 } from '@siafoundation/units'
 import BigNumber from 'bignumber.js'
 import {
@@ -173,10 +171,7 @@ export function transformDownGouging({
   }
 }
 
-export function transformDownPricePinning(
-  p: PricePinSettings,
-  periodBlocks?: number
-): PricePinData {
+export function transformDownPricePinning(p: PricePinSettings): PricePinData {
   const fixedFiat = currencyOptions.find((c) => c.id === p.currency)?.fixed || 6
   return {
     pinningEnabled: p.enabled,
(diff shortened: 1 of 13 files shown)


alexfreska:
* **#685** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/685?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/renterd/contexts/config/transformDown.ts
```diff
@@ -186,38 +181,20 @@ export function transformDownPricePinning(
...
-        periodBlocks || weeksToBlocks(6)
-      ),
-      fixedFiat
-    ),
-    shouldPinMaxRPCPrice: p.gougingSettingsPins?.maxRPCPrice.pinned,
-    maxRPCPriceMillionPinned: toFixedMaxBigNumber(
-      valuePerOneToPerMillion(
-        new BigNumber(p.gougingSettingsPins.maxRPCPrice.value)
-      ),
+      new BigNumber(p.autopilots['autopilot']?.allowance.value || 0),
```
alexfreska, line 184: @ChrisSchinnerl could you confirm:

1. The `autopilots.autopilot.allowance.value` is per month?

2. The UI currently converts autopilot allowance, download estimate, and upload estimates, from per period to per month for display - is it correct that the daemon uses these as per period?

ChrisSchinnerl: The allowance is per period. So if you pin that value to $10 and then later change the duration of a period, you will still only spend $10 per period. So the UI needs to convert that value from $ per period to $ per month itself. 

Upload and download is per TiB uploaded/downloaded
Storage is TB / month (30*144 blocks).
So those fields shouldn't require conversion on your end unless you want to display the bandwidth as TB instead of TiB.

ChrisSchinnerl (commented):
I think only thing missing is the allowance conversion but the rest looks good.

alexfreska:
> I think only thing missing is the allowance conversion but the rest looks good.

@ChrisSchinnerl ok should be good now 👍🏼

apps/renterd/contexts/config/transformDown.ts
ChrisSchinnerl: Oh yeah that's the expected upload and download per period

ChrisSchinnerl (approved):

PR #698, opened 2024-08-26
add explored api sdk to config, for use in future siacentral phase out

telestrial:
I figured this one conceptually made sense as its own little PR as opposed to tacking it onto the first phase out route PR and having it get lost in the history a bit.

The zen address is probably wrong here or there just isn't support for this yet. Should I remove that line or is there a current working address? Am I correct that we're using testnet-zen in a docker config to stand in for that local `index.ts` to switch between the different modes?

### DIFF
--- a/apps/explorer/config/index.ts
+++ b/apps/explorer/config/index.ts
@@ -10,3 +10,4 @@ export const isMainnet = true
 // APIs
 export const faucetApi = 'https://api.siascan.com/zen/faucet'
 export const siaCentralApi = 'https://api.siacentral.com/v2'
+export const exploredApi = 'https://explored.siascan.sia.dev/api'
--- a/apps/explorer/config/testnet-zen.ts
+++ b/apps/explorer/config/testnet-zen.ts
@@ -10,3 +10,4 @@ export const isMainnet = false
 // APIs
 export const faucetApi = 'https://api.siascan.com/zen/faucet'
 export const siaCentralApi = 'https://api.siacentral.com/v2/zen'
+export const exploredApi = 'https://explored.zen.siascan.sia.dev/api'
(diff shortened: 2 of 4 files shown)


apps/explorer/config/index.ts
```diff
@@ -10,3 +10,4 @@ export const isMainnet = true
 // APIs
 export const faucetApi = 'https://api.siascan.com/zen/faucet'
 export const siaCentralApi = 'https://api.siacentral.com/v2'
+export const exploredApi = 'https://explored.test.sia.tools/api'
```
alexfreska, line 13: For this use: `https://explored.siascan.sia.dev/api`

apps/explorer/config/testnet-zen.ts
```diff
@@ -10,3 +10,4 @@ export const isMainnet = false
 // APIs
 export const faucetApi = 'https://api.siascan.com/zen/faucet'
 export const siaCentralApi = 'https://api.siacentral.com/v2/zen'
+export const exploredApi = 'https://explored.test.sia.tools/zen/api' // Zen support may not exist at the time of this comment and/or this address may not be correct nor final.
```
alexfreska, line 13: For this use: `https://explored.zen.siascan.sia.dev/api`

alexfreska:
> Am I correct that we're using testnet-zen in a docker config to stand in for that local index.ts to switch between the different modes?

@telestrial not exactly, the config/index.ts file is swapped with config/testnet-zen.ts using an Nx file replacement during the NX build process. You can see this in the build targets: https://github.com/SiaFoundation/web/blob/cf995040d99acb6a538bdfc9ce42c6c2ed3ceec6/apps/explorer/project.json#L37-L44

alexfreska (approved):

PR #703, opened 2024-08-27
fix: renterd no recommendations

alexfreska:
- The recommendations now say "no recommendations" instead of 0. https://github.com/SiaFoundation/desktop/issues/90

### DIFF
--- a/apps/renterd/components/Config/Recommendations.tsx
+++ b/apps/renterd/components/Config/Recommendations.tsx
@@ -177,7 +177,9 @@ export function Recommendations() {
             {usableHostsCurrent}/{hostTarget50}
           </Text>
           <Text size="16" weight="medium">
-            {recommendations.length === 1
+            {recommendations.length === 0
+              ? 'No recommendations'
+              : recommendations.length === 1
               ? '1 recommendation'
               : `${recommendations.length} recommendations`}{' '}
             to match with more hosts
--- a/.changeset/small-cameras-lay.md
+++ b/.changeset/small-cameras-lay.md
@@ -1,0 +1,5 @@
+---
+'renterd': minor
+---
+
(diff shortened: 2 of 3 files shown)


alexfreska:
* **#703** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/703?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/renterd/components/Config/Recommendations.tsx
```diff
@@ -177,7 +177,9 @@ export function Recommendations() {
             {usableHostsCurrent}/{hostTarget50}
           </Text>
           <Text size="16" weight="medium">
-            {recommendations.length === 1
+            {recommendations.length === 0
+              ? 'No recommendations'
+              : recommendations.length === 1
               ? '1 recommendation'
               : `${recommendations.length} recommendations`}{' '}
```
telestrial, line 184: Here's another option that avoids the double ternary, in case you hadn't considered it:

```tsx
  {recommendations.length === 0 ? 'No recommendations' : `${recommendations.length} recommendation${recommendations.length > 1 ? 's' : ''}`}
```

You could also pull that example out into a `const` for readability, but what's above works well too! Approving regardless--just thought I'd share a different approach.

telestrial (approved):

apps/renterd/components/Config/Recommendations.tsx
alexfreska: @telestrial agreed that this could be cleaned up, I think the full solution is to add a really slick pluralize helper to use in all cases like this, follow up PR: https://github.com/SiaFoundation/web/pull/706

PR #710, opened 2024-08-30
refactor: renterd tests cluster

alexfreska:
- Add support for clusterd testing.
- Migrate all renterd tests to clusterd.
- Testing is now parallel, worker count: half logical cpus locally, 2 workers on CI
- Re-enabled testing against all browsers.
- Updated all go versions to 1.23.0, to align with cluster deps.

PR #712, opened 2024-08-30
fix: pluralize add tests

alexfreska:
- Fixed a bug in the pluralize helper, added tests.

### DIFF
--- a/.changeset/chatty-actors-shop.md
+++ b/.changeset/chatty-actors-shop.md
@@ -1,0 +1,5 @@
+---
+'@siafoundation/units': patch
+---
+
+Fixed a bug in the pluralize helper.
--- a/apps/renterd/components/Config/Recommendations.tsx
+++ b/apps/renterd/components/Config/Recommendations.tsx
@@ -179,7 +179,7 @@ export function Recommendations() {
           </Text>
           <Text size="16" weight="medium">
             {pluralize(recommendations.length, 'recommendation', {
-              zeroForm: 'No recommendations',
+              customZero: 'No recommendations',
             })}{' '}
             to match with more hosts
           </Text>
--- a/libs/units/src/pluralize.spec.ts
+++ b/libs/units/src/pluralize.spec.ts
@@ -1,0 +1,26 @@
+import { pluralize } from './pluralize'
+
+test('pluralize', () => {
+  expect(pluralize(0, 'test')).toBe('0 tests')
+  expect(pluralize(1, 'test')).toBe('1 test')
+  expect(pluralize(2, 'test')).toBe('2 tests')
+  expect(pluralize(0, 'child', 'children')).toBe('0 children')
+  expect(pluralize(1, 'child', 'children')).toBe('1 child')
+  expect(pluralize(2, 'child', 'children')).toBe('2 children')
+  expect(
+    pluralize(0, 'child', {
+      customZero: 'no children',
+    })
+  ).toBe('no children')
+  expect(
+    pluralize(3, 'child', {
+      customZero: 'no children',
+    })
+  ).toBe('3 childs')
+  expect(
+    pluralize(3, 'child', {
+      customZero: 'no children',
+      plural: 'children',
+    })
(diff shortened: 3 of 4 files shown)


alexfreska:
* **#710** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/710?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#712** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/712?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

telestrial:
@alexfreska Is this change mainly getting at this "children" example? What was the original undesirable behavior?

alexfreska:
> @alexfreska Is this change mainly getting at this "children" example? What was the original undesirable behavior?

@telestrial not specifically, I just added exhaustive tests for all usage patterns. It actually returned a malformed string for 3 of  the test cases, so just generally not correct. I "checked" it in the app, should have had these tests from the start. Results with the old method:
Screenshot 2024-08-30 at 1 50 30 PM

telestrial:
> It actually returned a malformed string for 3 of the test cases, so just generally not correct.

Gotcha. Yeah. I think this is a matter of scope: how comprehensive should this be? If we want an all-purpose pluralization function to use freely anywhere and everywhere, it may make sense to use an npm library. There are so so many edge cases and novel rules about pluralization in the English language. Words that add "s" or "es", ones that do neither, ones that just use a different ending altogether, and more. It would probably take a pretty large engineering effort to cover them all. If instead we're looking to cover this vertical of pluralization for specific words in our apps--hosts, renters, etc, it may make sense to do something a bit more explicit: a map of our known needed words and a function that handles that data structure. 

For the comprehensive route: I've used [this](https://github.com/plurals/pluralize) to some success in other projects. 

Even that library won't cover everything, though. This is a deceivingly complex problem in the English language.

telestrial (approved):

alexfreska:
> > It actually returned a malformed string for 3 of the test cases, so just generally not correct.
> 
> Gotcha. Yeah. I think this is a matter of scope: how comprehensive should this be? If we want an all-purpose pluralization function to use freely anywhere and everywhere, it may make sense to use an npm library. There are so so many edge cases and novel rules about pluralization in the English language. Words that add "s" or "es", ones that do neither, ones that just use a different ending altogether, and more. It would probably take a pretty large engineering effort to cover them all. If instead we're looking to cover this vertical of pluralization for specific words in our apps--hosts, renters, etc, it may make sense to do something a bit more explicit: a map of our known needed words and a function that handles that data structure.
> 
> For the comprehensive route: I've used [this](https://github.com/plurals/pluralize) to some success in other projects.
> 
> Even that library won't cover everything, though. This is a deceivingly complex problem in the English language.

I wanted to avoid pulling in a dependency, especially one with a dictionary - this API circumvents the complexity by simply having the caller provide the singular and plural words. The benefit being it abstracts all the nested ternary stuff.

telestrial:
> I wanted to avoid pulling in a dependency, especially one with a dictionary - this API circumvents the complexity by simply having the caller provide the singular and plural words. The benefit being it abstracts all the nested ternary stuff.

Totally agree on the library. Just outlining options, even the less great ones. I think your solution should work well but we may find edge cases over time that challenge the parameters here. Maybe not, though! There's so many weird rules to remember/know them all.

alexfreska:
> > I wanted to avoid pulling in a dependency, especially one with a dictionary - this API circumvents the complexity by simply having the caller provide the singular and plural words. The benefit being it abstracts all the nested ternary stuff.
> 
> Totally agree on the library. Just outlining options, even the less great ones. I think your solution should work well but we may find edge cases over time that challenge the parameters here. Maybe not, though! There's so many weird rules to remember/know them all.

@telestrial 👍🏼 yep we may need to adjust the options as we go. Oddly I couldn't find thaaat many places we needed pluralization - maybe just my grepping didn't catch em.

PR #725, opened 2024-09-05
Use working directory instead of package for go test

n8maninger:
Since we don't have a `go.work` file any more, `go test` needs to use the package working directory. Should fix all this
Screenshot 2024-09-05 at 11 51 57 AM


alexfreska (approved):
👍🏼

PR #728, opened 2024-09-06
fix(design-system): force the ThemeRadio component to render only on the client side

telestrial:
This avoids the hydration error/warning we were receiving from the ThemeRadio component, concerning the clash between a client's localStorage theme preference and a server's lack of localStorage.

### DIFF
--- a/.changeset/selfish-mice-relax.md
+++ b/.changeset/selfish-mice-relax.md
@@ -1,0 +1,5 @@
+---
+'@siafoundation/design-system': minor
+---
+
+A wrapping ClientSideOnly component has been created that forces its children to render only on the client side.
--- a/.changeset/unlucky-lobsters-reply.md
+++ b/.changeset/unlucky-lobsters-reply.md
@@ -1,0 +1,5 @@
+---
+'explorer': minor
+---
+
+The light and dark ThemeRadio footer component now renders only on the client side.
--- a/apps/explorer/components/Layout/Footer.tsx
+++ b/apps/explorer/components/Layout/Footer.tsx
@@ -5,6 +5,7 @@ import {
   Logo,
   ThemeRadio,
   CurrencyFiatSelector,
+  ClientSideOnly,
 } from '@siafoundation/design-system'
 
 export function Footer() {
@@ -49,7 +50,11 @@ export function Footer() {
           <div className="flex-1" />
           <div className="flex-1 flex items-center justify-end gap-6">
             <CurrencyFiatSelector />
-            <ThemeRadio className="hidden md:flex" />
+            <ClientSideOnly
+              fallback={<div className="w-[80px] h-[16px]"></div>}
+            >
+              <ThemeRadio className="hidden md:flex" />
+            </ClientSideOnly>
           </div>
         </div>
       </div>
--- a/libs/design-system/src/components/ClientSideOnly.tsx
+++ b/libs/design-system/src/components/ClientSideOnly.tsx
@@ -1,0 +1,24 @@
+'use client'
+
+import { useEffect, useState } from 'react'
+
+/**
+ * A wrapping component that forces its children to render on the client side.
+ */
+export function ClientSideOnly({
+  children,
+  fallback,
+}: {
+  children: React.ReactNode
+  fallback?: React.ReactNode
+}) {
+  const [shouldRender, setShouldRender] = useState(false)
+
+  useEffect(() => {
+    setShouldRender(true)
+  }, [])
+
+  if (!shouldRender) return fallback || null
+
+  return children
+}
--- a/libs/design-system/src/components/ThemeRadio.tsx
+++ b/libs/design-system/src/components/ThemeRadio.tsx
@@ -32,6 +32,7 @@ type Props = {
 
 export function ThemeRadio({ className, tooltipClassName, tabIndex }: Props) {
   const { theme, setTheme } = useTheme()
+
   return (
     <RadioGroupPrimitive.Root
       value={theme}
--- a/libs/design-system/src/index.ts
+++ b/libs/design-system/src/index.ts
@@ -76,6 +76,7 @@ export * from './components/PaginatorKnownTotal'
 export * from './components/PaginatorUnknownTotal'
 export * from './components/PaginatorMarker'
 export * from './components/ListWithSeparators'
+export * from './components/ClientSideOnly'
 
 // app
 export * from './app/AppPublicLayout'


telestrial:
* **#728** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/728?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

libs/design-system/src/components/ThemeRadio.tsx
```diff
@@ -31,7 +31,22 @@ type Props = {
 }
 
 export function ThemeRadio({ className, tooltipClassName, tabIndex }: Props) {
+  const [shouldMount, setShouldMount] = useState(false)
   const { theme, setTheme } = useTheme()
+
+  // This effect, the shouldMount state above, and the if check below avoids
+  // an unintuitive hydration warning with NextJS by making sure this component
+  // *only* renders on the client side, as we intend with the 'use client'
```
alexfreska, line 39: Map of rendering behaviour:
- no directive: server-only
- 'use client': runs on both client and server (unintuitive name)
- your change: client-only


The daemon apps all ship as client-side only apps since they have to be exported and served as static files. Their root component has a similar check. We can use this existing component:
https://github.com/SiaFoundation/web/blob/1cd992a28b2128f158bc04ec37045b48d49556d5/libs/design-system/src/site/NextAppCsr.tsx#L9-L10
The benefit beyond reuse being the intent becomes very clear when you see something wrapped with `<ClientSideOnly>`.

We should move it into its own file in src/components, remove the wrapping div, and I think renaming it `ClientSideOnly` would be more clear.

alexfreska (commented):
Nice fix 👍🏼 👍🏼 Left one comment about extracting the behaviour into existing component.

libs/design-system/src/components/ThemeRadio.tsx
telestrial: I have pushed changes to reflect this. I did have one thought while doing it, though, that is true for the local solution I had and this new implementation you suggested:

We will run into cumulative layout shift issues down the road with this, and the wrapping nature of this component may actually make it tougher to combat. I think this works fine for now, but, in the future, we may want to actually make this a hook that serves up "shouldMount" and then lets the component decide what to do in the false case.  This ThemeRadio component is actually a great example. In the footer, we see the USD dropdown start from right and then get pushed out to left. With a hook like I'm thinking about, the ThemeRadio component could invoke this and then have the responsibility of managing its own placeholder. I can't think of other ways to do it, but there certainly may be.

.changeset/unlucky-lobsters-reply.md
```diff
@@ -0,0 +1,5 @@
+---
+'explorer': minor
+---
+
+The light and dark Themeradio footer component now renders only on the client side.
```
alexfreska, line 5: ThemeRadio

.changeset/selfish-mice-relax.md
```diff
@@ -0,0 +1,5 @@
+---
+'@siafoundation/design-system': minor
+---
+
+A wrapping ClientSideOnly component has been created that forcers its chidlren to render only on the client side.
```
alexfreska, line 5: forcers
chidlren

libs/design-system/src/components/ThemeRadio.tsx
alexfreska: You could add a fallback prop that is rendered instead of null.

libs/design-system/src/components/ClientSideOnly.tsx
```diff
@@ -0,0 +1,18 @@
...
+ * A wrapping component that forces its children to render on the client side.
+ */
+export function ClientSideOnly({ children }: { children: React.ReactNode }) {
+  const [shouldRender, setShouldRender] = useState(false)
+
+  useEffect(() => {
+    setShouldRender(true)
+  }, [])
+
+  if (!shouldRender) return null
```
alexfreska, line 15: Relating to your other comment, `fallback` could be rendered here:
`if (!shouldRender) return fallback || null`

apps/explorer/components/Layout/Footer.tsx
```diff
@@ -49,7 +50,9 @@ export function Footer() {
           <div className="flex-1" />
           <div className="flex-1 flex items-center justify-end gap-6">
             <CurrencyFiatSelector />
-            <ThemeRadio className="hidden md:flex" />
+            <ClientSideOnly>
```
alexfreska, line 53: This will always be necessary for ThemeRadio so you can move it inside that component.

libs/design-system/src/components/ThemeRadio.tsx
```diff
@@ -32,6 +32,7 @@ type Props = {
 
 export function ThemeRadio({ className, tooltipClassName, tabIndex }: Props) {
   const { theme, setTheme } = useTheme()
+
```
alexfreska, line 35: Relating to your other comment, to avoid the layout shift you could render the whole thing with a fixed value in fallback so that it looks the same/takes the same exact space, and also does not hydration mismatch.
```
return <ClientSideOnly fallback={<ThemeRadioInner value={false} ... />}>
  <ThemeRadioComponent value={theme} ... />
  </ClientSideOnly>
  ```

apps/explorer/components/Layout/Footer.tsx
telestrial: Happy to do this but it's worth noting that it isn't always necessary for this component. We use this in Explorer's menu and NextJS isn't hydrating that version of the component because it isn't visible. You can test this out by getting the hydration warning on main and then removing the footer ThemeRadio. That particular warning will go away.

alexfreska: I see what you are saying, but this requires the caller to understand the implementation details - with this logic internalized the component is safe to use however.

alexfreska: The other added benefit is that you can internally set the fallback to look exactly the same rather than externally set a equivalent sized box. If someone ever changes ThemeRadio that box (and fallbacks in other apps) may fall out of sync.

alexfreska (approved):
Left some additional comments on this thread: https://github.com/SiaFoundation/web/pull/728#discussion_r1750666312 - you can make the final call on whether to adjust that, otherwise looks good 👍🏼

telestrial:
### Merge activity

* **Sep 9, 4:22 PM EDT**: @telestrial started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/SiaFoundation/web/728).

PR #747, opened 2024-09-20
refactor(renterd): typescript strict mode for config feature

alexfreska:
> Strict mode refactoring part 3/3: no changes to functionality.

- Update renterd app to TypeScript strict mode 2/2: config feature.
- Primarily adds all missing strict null checks.
- Refine config data types to be more specific.

### DIFF
--- a/apps/renterd/contexts/config/deriveAllowance.ts
+++ b/apps/renterd/contexts/config/deriveAllowance.ts
@@ -50,17 +50,17 @@ export function derivePricingFromAllowance({
   downloadWeight = 5,
   uploadWeight = 1,
 }: {
-  allowanceMonth: BigNumber
+  allowanceMonth?: BigNumber
   allowanceFactor?: number
-  storageTB: BigNumber
-  downloadTBMonth: BigNumber
-  uploadTBMonth: BigNumber
-  redundancyMultiplier: BigNumber
+  storageTB?: BigNumber
+  downloadTBMonth?: BigNumber
+  uploadTBMonth?: BigNumber
+  redundancyMultiplier?: BigNumber
   storageWeight?: number
   downloadWeight?: number
   uploadWeight?: number
 }) {
-  // Return null if zero values are provided.
+  // Return undefined if zero values are provided.
   if (
     !allowanceMonth?.gt(0) ||
     allowanceFactor <= 0 ||
@@ -69,7 +69,7 @@ export function derivePricingFromAllowance({
     !downloadTBMonth?.gt(0) ||
     !uploadTBMonth?.gt(0)
   ) {
-    return null
+    return undefined
   }
   // Apply scaling factor to allowance.
   const scaledAllowance = allowanceMonth.times(allowanceFactor)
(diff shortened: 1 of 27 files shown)


alexfreska:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/747?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#747** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/747?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#746** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/746?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

apps/renterd/contexts/config/deriveAllowance.ts
```diff
@@ -50,17 +50,17 @@ export function derivePricingFromAllowance({
...
+  storageTB?: BigNumber
+  downloadTBMonth?: BigNumber
+  uploadTBMonth?: BigNumber
+  redundancyMultiplier?: BigNumber
   storageWeight?: number
   downloadWeight?: number
   uploadWeight?: number
 }) {
-  // Return null if zero values are provided.
+  // Return undefined if zero values are provided.
```
peterjan, line 63: `nit` would be easier on the eyes if we would assert they're zero instead of not greater than because they can't be negative right

apps/renterd/contexts/config/useRedundancyMultiplier.tsx
```diff
@@ -6,8 +6,8 @@ export function useRedundancyMultiplier({
   minShards,
   totalShards,
 }: {
-  minShards: BigNumber
-  totalShards: BigNumber
+  minShards?: BigNumber
```
peterjan, line 9: I understand most of the changes but this one looks a bit off. Why would we want to allow optional parameters here?

peterjan (approved):
I trust this is an improvement 👍 😂 I understand what you are doing here but it was my understanding that `null` is more explicit than `undefined`. So in some cases I don't see how going from null to undefined is making it more strict.

apps/renterd/contexts/config/useRedundancyMultiplier.tsx
alexfreska: This method and many others have to have optional params to handle when the user clears the field, or if the data is still loading from the server, both of which would result in undefined values - it may be more accurate to make this minShards: BigNumber|undefined meaning the caller has to pass/provide the param, but it can be undefined.

alexfreska:
@peterjan 
> I trust this is an improvement 👍 😂 I understand what you are doing here but it was my understanding that `null` is more explicit than `undefined`. So in some cases I don't see how going from null to undefined is making it more strict.

With the strict null checks  null and undefined are treated as distinct types, so `null` no longer works when passed into an optional type constraint like foo?: string, it has to be `undefined`.

alexfreska:
### Merge activity

* **Sep 30, 12:42 PM EDT**: @alexfreska started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/SiaFoundation/web/747).

PR #760, opened 2024-10-01
feat(explored-types): re-export base SiacoinOutput type

telestrial:

PR #759, opened 2024-10-01
feat(explored): add contractRevisions APIs

telestrial:

PR #761, opened 2024-10-01
feat(explorer): use explored for the contracts route

telestrial:
## Questions/Points of Concern (Updated 10/23)

* The exact timestamp for negotiation, expiration, proof deadline, and payout are a bit off from Sia Central, but I believe that's expected. We're doing this on the client side now with respective block heights. The difference is not significant but I don't know how exact these are expected to be. `blockHeightToHumanDate` in `lib/contracts` would be the relevant pipe they are all flowing through.

* Payout is also a little different, too. I've never seen more that a few cents, but wanted to call that out, too. As far as I know, that's a difference on the explored side.

* Do the resolved from/to buttons work as we would expect? I believe so but it's hard to find good test contracts that have both. I've only see one or the other. It shouldn't matter, but I like to visually/manually test as much as I can and couldn't find multiple contract renewals.

* Does the status make sense (in progress, obligation successful, obligation failed)? This is another one where I'd ideally find one of every kind but that's a little more detailed understanding than I may have right now.

* The network calls at the top of the Contract next page, their order, and their error handling.I may not have them all ordered the most optimal way and then: are we only returning `notFound()` on an absent contract record? There may be a better way to do all of this, but we have multiple chained calls that rely on previous.

Comparing to a contract page through Sia Central, we're very close, I think. ~~I've left two review comments open down below with two other things to discuss.~~

### DIFF
--- a/apps/explorer/app/contract/[id]/page.tsx
+++ b/apps/explorer/app/contract/[id]/page.tsx
@@ -1,4 +1,3 @@
-import { SiaCentralContract } from '@siafoundation/sia-central-types'
 import { ContractView } from '../../../components/ContractView'
 import { Metadata } from 'next'
 import { routes } from '../../../config/routes'
@@ -8,6 +7,7 @@ import { notFound } from 'next/navigation'
 import { stripPrefix, truncate } from '@siafoundation/design-system'
 import { to } from '@siafoundation/request'
 import { explored } from '../../../config/explored'
+import { ChainIndex, ExplorerFileContract } from '@siafoundation/explored-types'
 
 export function generateMetadata({ params }): Metadata {
   const id = decodeURIComponent((params?.id as string) || '')
@@ -24,15 +24,15 @@ export function generateMetadata({ params }): Metadata {
 export const revalidate = 0
 
 export default async function Page({ params }) {
-  const id = params?.id as string
-  const [[c, error], [r]] = await Promise.all([
-    to(
-      siaCentral.contract({
-        params: {
-          id,
-        },
-      })
-    ),
+  const id = params?.id
+
+  // Grab the contract and previous revisions data.
+  const [
+    [rate, rateError],
+    [contract, contractError],
+    [previousRevisions, previousRevisionsError],
+    [currentTip, currentTipError],
+  ] = await Promise.all([
     to(
       siaCentral.exchangeRates({
         params: {
@@ -40,100 +40,99 @@ export default async function Page({ params }) {
         },
       })
     ),
-  ])
-
-  if (error) {
-    throw error
-  }
-
-  const contract = c?.contract
-
-  if (!contract) {
-    return notFound()
-  }
-
-  const formationTxnId = getFormationTxnId(contract)
-  const finalRevisionTxnId = contract?.transaction_id || ''
-
-  const [[ft], [rt]] = await Promise.all([
-    to(
-      siaCentral.transaction({
-        params: {
-          id: formationTxnId,
-        },
-      })
-    ),
-    to(
-      siaCentral.transaction({
-        params: {
-          id: finalRevisionTxnId,
-        },
-      })
-    ),
-  ])
-
-  const formationTransaction = ft?.transaction
-  const renewedFrom = formationTransaction?.contract_revisions?.[0]
-  const renewalTransaction = rt?.transaction
-  const renewedTo = renewalTransaction?.storage_contracts?.[0]
-
-  // The following is a temporary addition to satisfy new Transaction
-  // component requirements for the Sia Central phase out on the
-  // transaction route.
-  const [
-    [transaction, transactionError],
-    [transactionChainIndices, transactionChainIndicesError],
-    [currentTip, currentTipError],
-  ] = await Promise.all([
-    to(
-      explored.transactionByID({
-        params: { id: formationTransaction?.id || '' },
-      })
-    ),
-    to(
-      explored.transactionChainIndices({
-        params: { id: formationTransaction?.id || '' },
-      })
-    ),
+    to(explored.contractByID({ params: { id } })),
+    to<ExplorerFileContract[]>(explored.contractRevisions({ params: { id } })),
     to(explored.consensusTip()),
   ])
 
-  if (transactionError) throw transactionError
-  if (transactionChainIndicesError) throw transactionChainIndicesError
+  if (rateError) throw rateError
+
+  if (contractError) throw contractError
+  if (!contract) return notFound()
+
+  if (previousRevisionsError) throw previousRevisionsError
+  if (!previousRevisions)
+    throw new Error('No previousRevisions found in successful request')
+
   if (currentTipError) throw currentTipError
-  if (!transaction || !transactionChainIndices || !currentTip) return notFound()
+  if (!currentTip) throw new Error('No currentTip found in successful request')
+
+  const formationTxnID =
+    previousRevisions[0].transactionID ?? contract.transactionID
+  const finalRevisionTxnID =
+    previousRevisions[previousRevisions.length - 1].transactionID ??
+    contract.transactionID
+
+  // Fetch our formation and finalRevision transaction information.
+  const [
+    [formationTransaction, formationTransactionError],
+    [renewalTransaction, renewalTransactionError],
+  ] = await Promise.all([
+    to(
+      explored.transactionByID({
+        params: {
+          id: formationTxnID,
+        },
+      })
+    ),
+    to(
+      explored.transactionByID({
+        params: {
+          id: finalRevisionTxnID,
+        },
+      })
+    ),
+  ])
+
+  if (formationTransactionError) throw formationTransactionError
+  if (!formationTransaction)
+    throw new Error('No formation transaction found in successful request')
+
+  if (renewalTransactionError) throw renewalTransactionError
+  if (!renewalTransaction)
+    throw new Error('No renewal transaction found in successful request')
+
+  const renewedFromID =
+    formationTransaction.fileContractRevisions?.[0].id ?? null
+  const renewedToID = renewalTransaction.fileContracts?.[0].id ?? null
+
+  const [formationTxnChainIndices, formationTxnChainIndicesError] = await to<
+    ChainIndex[]
+  >(
+    explored.transactionChainIndices({
+      params: { id: formationTransaction.id },
+    })
+  )
+
+  if (formationTxnChainIndicesError) throw formationTxnChainIndicesError
+  if (!formationTxnChainIndices)
+    throw new Error('No formationTxnChainIndices found in successful request')
 
   // Use the first chainIndex from the above call to get our parent block.
   const [parentBlock, parentBlockError] = await to(
-    explored.blockByID({ params: { id: transactionChainIndices[0].id } })
+    explored.blockByID({ params: { id: formationTxnChainIndices[0].id } })
   )
 
   if (parentBlockError) throw parentBlockError
-  if (!parentBlock) return notFound()
+  if (!parentBlock)
+    throw new Error('No parentBlock found in successful request')
 
   return (
     <ContractView
+      previousRevisions={previousRevisions}
+      currentHeight={currentTip.height}
       contract={contract}
-      rates={r?.rates}
-      renewedFrom={renewedFrom}
-      renewedTo={renewedTo}
-      formationTransaction={transaction}
+      rates={rate?.rates}
+      renewedFromID={renewedFromID ? stripPrefix(renewedFromID) : renewedFromID}
+      renewedToID={renewedToID ? stripPrefix(renewedToID) : renewedToID}
+      formationTransaction={formationTransaction}
+      formationTxnChainIndex={formationTxnChainIndices}
       formationTransactionHeaderData={{
-        id: stripPrefix(transaction.id),
-        blockHeight: transactionChainIndices[0].height,
-        confirmations: currentTip.height - transactionChainIndices[0].height,
+        id: stripPrefix(formationTransaction.id),
+        blockHeight: formationTxnChainIndices[0].height,
+        confirmations: currentTip.height - formationTxnChainIndices[0].height,
         timestamp: parentBlock.timestamp,
       }}
     />
   )
-}
-
-function getFormationTxnId(contract: SiaCentralContract) {
-  let id = contract?.transaction_id
-  if (contract?.previous_revisions?.length) {
-    id =
-      contract.previous_revisions[contract.previous_revisions?.length - 1]
-        .transaction_id
-  }
-  return id
 }
--- a/apps/explorer/components/Contract/ContractHeader.tsx
+++ b/apps/explorer/components/Contract/ContractHeader.tsx
@@ -2,46 +2,46 @@ import {
   Badge,
   ContractTimeline,
   LinkButton,
+  stripPrefix,
 } from '@siafoundation/design-system'
 import { ArrowLeft16, ArrowRight16 } from '@siafoundation/react-icons'
-import { SiaCentralContract } from '@siafoundation/sia-central-types'
-import { lowerCase } from '@technically/lodash'
 import { routes } from '../../config/routes'
 import { EntityHeading } from '../EntityHeading'
-import { siaCentral } from '../../config/siaCentral'
-import { to } from '@siafoundation/request'
+import { ChainIndex, ExplorerFileContract } from '@siafoundation/explored-types'
+import { determineContractStatus } from '../../lib/contracts'
 
 type Props = {
-  contract: SiaCentralContract
-  renewedToId?: string
-  renewedFromId?: string
+  currentHeight: number
+  contract: ExplorerFileContract
+  renewedToID: string | null
+  renewedFromID: string | null
+  formationTxnChainIndex: ChainIndex[]
 }
 
 export async function ContractHeader({
+  currentHeight,
   contract,
-  renewedFromId,
-  renewedToId,
+  renewedFromID,
+  renewedToID,
+  formationTxnChainIndex,
 }: Props) {
   const { id } = contract
-  const [latest, error] = await to(siaCentral.blockLatest())
-  if (error) {
-    console.error(error.stack)
-  }
+  const contractStatus = determineContractStatus(contract)
   return (
     <div className="flex flex-col gap-x-4 gap-y-4 pb-4">
       <div className="flex flex-wrap gap-x-4 gap-y-4 items-center justify-between">
         <EntityHeading
           label="Contract"
           type="contract"
-          value={id}
+          value={stripPrefix(id)}
           href={routes.contract.view.replace(':id', id)}
         />
         <div className="flex gap-1">
-          {renewedFromId && renewedFromId !== id && (
+          {renewedFromID && renewedFromID !== stripPrefix(id) && (
             <LinkButton
               className="hidden sm:flex"
               variant="gray"
-              href={routes.contract.view.replace(':id', renewedFromId)}
+              href={routes.contract.view.replace(':id', renewedFromID)}
               data-testid="explorer-contract-renewedFrom"
             >
               <ArrowLeft16 />
@@ -51,15 +51,19 @@ export async function ContractHeader({
           <Badge
             interactive={false}
             variant={
-              contract.status === 'obligationSucceeded' ? 'green' : 'gray'
+              contractStatus === 'in progress'
+                ? 'gray'
+                : contractStatus === 'obligation succeeded'
+                ? 'green'
+                : 'red'
             }
           >
-            {lowerCase(contract.status)}
+            {contractStatus}
           </Badge>
-          {renewedToId && renewedToId !== id && (
+          {renewedToID && renewedToID !== stripPrefix(id) && (
             <LinkButton
               className="hidden sm:flex"
-              href={routes.contract.view.replace(':id', renewedToId)}
+              href={routes.contract.view.replace(':id', renewedToID)}
               data-testid="explorer-contract-renewedTo"
             >
               renewed to
@@ -68,22 +72,29 @@ export async function ContractHeader({
           )}
         </div>
       </div>
-      {latest?.block && (
+      {currentHeight && (
         <div className="px-1">
           <ContractTimeline
-            currentHeight={latest.block.height || 0}
-            contractHeightStart={contract.negotiation_height}
-            contractHeightEnd={contract.expiration_height}
-            proofWindowHeightStart={contract.expiration_height}
-            proofWindowHeightEnd={contract.proof_deadline}
-            proofHeight={contract.proof_height}
+            currentHeight={currentHeight || 0}
+            contractHeightStart={
+              contract.confirmationIndex?.height ||
+              formationTxnChainIndex[0].height
+            }
+            contractHeightEnd={contract.windowStart}
+            proofWindowHeightStart={contract.windowStart}
+            proofWindowHeightEnd={contract.windowEnd}
+            proofHeight={contract.proofIndex ? contract.proofIndex.height : 0}
             range={{
-              startHeight: contract.negotiation_height,
+              startHeight:
+                contract.confirmationIndex?.height ||
+                formationTxnChainIndex[0].height,
               endHeight: Math.max(
-                latest.block.height || 0,
+                currentHeight || 0,
                 Math.round(
-                  contract.proof_deadline +
-                    (contract.expiration_height - contract.negotiation_height)
+                  contract.windowEnd +
+                    (contract.windowStart -
+                      (contract.confirmationIndex?.height ||
+                        formationTxnChainIndex[0].height))
                 )
               ),
             }}
--- a/apps/explorer/components/Contract/index.tsx
+++ b/apps/explorer/components/Contract/index.tsx
@@ -2,34 +2,48 @@
 
 import { useMemo } from 'react'
 import BigNumber from 'bignumber.js'
-import {
-  SiaCentralContract,
-  SiaCentralExchangeRates,
-  SiaCentralPartialSiacoinOutput,
-} from '@siafoundation/sia-central-types'
-import { humanBytes, humanDate, humanSiacoin } from '@siafoundation/units'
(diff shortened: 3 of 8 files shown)


telestrial:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/761?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#761** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/761?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#760** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/760?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#759** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/759?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

apps/explorer/app/contract/[id]/page.tsx
```diff
@@ -42,44 +38,58 @@ export default async function Page({ params }) {
     ),
   ])
 
-  if (error) {
-    throw error
-  }
+  if (rateError) throw rateError
+  if (!rate) return notFound()
```
Comment (alexfreska) [inline apps/explorer/app/contract/[id]/page.tsx:42]: This should probably throw an error so the error page is rendered rather than 404.

apps/explorer/app/contract/[id]/page.tsx
```diff
@@ -42,44 +38,58 @@ export default async function Page({ params }) {
...
 
-  const contract = c?.contract
+  // Grab the contract and previous revisions data.
```
Comment (alexfreska) [inline apps/explorer/app/contract/[id]/page.tsx:44]: Why not parallel with the rates?

apps/explorer/app/contract/[id]/page.tsx
```diff
@@ -42,44 +38,58 @@ export default async function Page({ params }) {
...
-  if (!contract) {
-    return notFound()
-  }
+  if (contractError) throw contractError
+  if (previousRevisionsError) throw previousRevisionsError
+  if (!contract || !previousRevisions) return notFound()
 
-  const formationTxnId = getFormationTxnId(contract)
-  const finalRevisionTxnId = contract?.transaction_id || ''
+  // What should we do here? What's going on now is wrong, I'm sure, and
```
Comment (alexfreska) [inline apps/explorer/app/contract/[id]/page.tsx:59]: The formation txn is the confirmation txn ID for the first ever revision.
The final revision txn ID is the confirmation txn ID for the last revision.
You already have the list of revisions, sort them by revision number.

apps/explorer/components/Contract/ContractHeader.tsx
```diff
@@ -50,10 +54,10 @@ export async function ContractHeader({
           <Badge
             interactive={false}
             variant={
-              contract.status === 'obligationSucceeded' ? 'green' : 'gray'
+              contract.valid && contract.resolved ? 'green' : 'gray' // This is an assumption about how this is derived.
```
alexfreska, line 57: This should be fine.

apps/explorer/components/Contract/ContractHeader.tsx
```diff
@@ -1,32 +1,36 @@
...
 type Props = {
-  contract: SiaCentralContract
+  currentHeight: number
+  contract: ExplorerFileContract
   renewedToId?: string
   renewedFromId?: string
 }
 
+// This is dummy logic and I'm not even sure it makes sense, given the values.
+function contractStatus(valid: boolean, resolved: boolean) {
```
alexfreska, line 19: Can be simplified to:


```suggestion
function contractStatus(valid: boolean, resolved: boolean) {
  return !valid ? 'invalid' : resolved ? 'resolved' : 'unresolved';
}
```

apps/explorer/components/Contract/ContractHeader.tsx
```diff
@@ -66,26 +70,27 @@ export async function ContractHeader({
...
         </div>
       </div>
-      {latest?.block && (
+      {currentHeight && (
         <div className="px-1">
-          <ContractTimeline
-            currentHeight={latest.block.height || 0}
+          {/* <ContractTimeline
+            currentHeight={currentHeight || 0}
             contractHeightStart={contract.negotiation_height}
```
alexfreska, line 77: `contractHeightStart={contract.confirmationIndex.height}`

apps/explorer/components/Contract/index.tsx
```diff
@@ -2,152 +2,162 @@
...
 type Props = {
-  contract: SiaCentralContract
+  previousRevisions: FileContractElement[]
+  currentHeight: number
+  contract: ExplorerFileContract
   rates?: SiaCentralExchangeRates
-  renewedTo?: SiaCentralContract
-  renewedFrom?: SiaCentralContract
+  renewedTo?: ExplorerFileContract
+  renewedFrom?: ExplorerFileContract
```
alexfreska, line 25: It looks like these props could be simplified to just the IDs.

apps/explorer/components/Contract/index.tsx
```diff
@@ -2,152 +2,162 @@
...
-        label: 'negotiation time',
-        copyable: false,
-        value:
-          contract.negotiation_timestamp !== '0001-01-01T00:00:00Z'
-            ? humanDate(contract.negotiation_timestamp, {
-                dateStyle: 'medium',
-                timeStyle: 'short',
-              })
-            : '-',
+        value: String(contract.valid),
```
alexfreska, line 66: This should be `!!contract.proofTransactionID`. Also, looking at the contract type, I am guessing all 3 proof related fields should be optional.

apps/explorer/components/Contract/index.tsx
```diff
@@ -2,152 +2,162 @@
...
       },
+      // {
+      //   label: 'negotiation height',
+      //   copyable: false,
+      //   value: contract.negotiation_height?.toLocaleString() || '-',
```
alexfreska, line 71: Use confirmation index for these fields. Get the time with `blockHeightToTime` method.

apps/explorer/components/Contract/index.tsx
```diff
@@ -2,152 +2,162 @@
...
+      //       ? humanDate(contract.expiration_timestamp, {
+      //           dateStyle: 'medium',
+      //           timeStyle: 'short',
+      //         })
+      //       : '-',
+      // },
+      // {
+      //   label: 'proof height',
+      //   copyable: false,
+      //   value: contract.proof_height
```
alexfreska, line 103: `contract.proofIndex.height`

apps/explorer/components/Contract/index.tsx
```diff
@@ -2,152 +2,162 @@
...
+      //     ? contract.proof_height.toLocaleString()
+      //     : '-',
+      // },
+      // {
+      //   label: 'proof time',
+      //   copyable: false,
+      //   value:
+      //     contract.proof_timestamp !== '0001-01-01T00:00:00Z'
+      //       ? humanDate(contract.proof_timestamp, {
```
alexfreska, line 112: `blockHeightToTime`

apps/explorer/components/Contract/index.tsx
```diff
@@ -157,14 +167,14 @@ export function Contract({ contract, rates, renewedFrom, renewedTo }: Props) {
         {
           label: 'renter payout: remaining renter allowance',
           initials: 'r',
-          sc: new BigNumber(renterPayoutValid.value),
-          hash: renterPayoutValid.output_id,
+          sc: renterPayoutValid && new BigNumber(renterPayoutValid.value),
+          hash: renterPayoutValid?.address,
```
alexfreska, line 171: output id and address are not the same, if we are changing what is displayed just make sure this is labeled correctly.

apps/explorer/components/Contract/index.tsx
```diff
@@ -274,39 +289,40 @@ export function Contract({ contract, rates, renewedFrom, renewedTo }: Props) {
...
     return false
   }
 
   const { renterPayoutValid, renterPayoutMissed } =
     getNewContractFormattedOutputs(contract)
 
   // renter payout valid and missed should be the same
-  if (renterPayoutValid.value !== renterPayoutMissed.value) {
+  if (renterPayoutValid?.value !== renterPayoutMissed?.value) {
+    // Does we need to catch undefined cases now?
```
alexfreska, line 307: I believe the earlier check does ensure these will be defined, although TypeScript is not smart enough to provide that here.

apps/explorer/components/Contract/index.tsx
```diff
@@ -318,48 +334,49 @@ function isProperlyFormedRenewedContract(contract: SiaCentralContract) {
...
-function getNewContractFormattedOutputs(contract: SiaCentralContract) {
+function getNewContractFormattedOutputs(contract: ExplorerFileContract) {
   return {
-    renterPayoutValid: contract.valid_proof_outputs[0],
-    renterPayoutMissed: contract.missed_proof_outputs[0],
-    hostPayoutValid: contract.valid_proof_outputs[1],
-    hostPayoutMissed: contract.missed_proof_outputs[1],
-    hostBurned: contract.missed_proof_outputs[2],
+    renterPayoutValid: contract.fileContract.validProofOutputs?.[0],
+    renterPayoutMissed: contract.fileContract.missedProofOutputs?.[0],
```
alexfreska, line 357: We should double check with the team that all these assumptions around index position are still valid. But can also get a good idea from comparisons to production siascan.com.

apps/explorer/components/Contract/index.tsx
```diff
@@ -2,152 +2,162 @@
...
+      //           dateStyle: 'medium',
+      //           timeStyle: 'short',
+      //         })
+      //       : '-',
+      // },
+      // {
+      //   label: 'payout height',
+      //   copyable: false,
+      //   value: contract.payout_height?.toLocaleString() || '-',
+      // },
```
alexfreska, line 138: This is how we calculate payout height in renterd:
```
  const payoutHeight =
    c.resolutionHeight > 0
      ? c.resolutionHeight + 144
      : c.revision.windowEnd + 144
```
We just first need to confirm how to get resolution height.

apps/explorer/app/contract/[id]/page.tsx
```diff
@@ -42,44 +38,58 @@ export default async function Page({ params }) {
...
     ),
   ])
 
-  const formationTransaction = ft?.transaction
-  const renewedFrom = formationTransaction?.contract_revisions?.[0]
-  const renewalTransaction = rt?.transaction
-  const renewedTo = renewalTransaction?.storage_contracts?.[0]
+  if (formationTransactionError) throw formationTransactionError
+  if (renewalTransactionError) throw renewalTransactionError
+  if (!formationTransaction || !renewalTransaction) return notFound()
```
Comment (alexfreska) [inline apps/explorer/app/contract/[id]/page.tsx:87]: An active contract will not be renewed yet, so we should not have this 404.

apps/explorer/app/contract/[id]/page.tsx
```diff
@@ -42,44 +38,58 @@ export default async function Page({ params }) {
...
+  const renewedFrom = formationTransaction.fileContractRevisions[0]
+  const renewedTo = renewalTransaction.fileContracts[0]
 
-  // The following is a temporary addition to satisfy new Transaction
-  // component requirements for the Sia Central phase out on the
-  // transaction route.
+  // Do our transaction fetching, chain indices, and tip fetching.
   const [
     [transaction, transactionError],
     [transactionChainIndices, transactionChainIndicesError],
```
Comment (alexfreska) [inline apps/explorer/app/contract/[id]/page.tsx:95]: <i class='graphite__hidden'>[Re: line +99]<br/><br/></i>

I believe this is refetching a txn we already have.<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/761?utm_source=unchanged-line-comment">Graphite</a>.</span>

apps/explorer/app/contract/[id]/page.tsx
```diff
@@ -101,6 +111,7 @@ export default async function Page({ params }) {
   if (transactionError) throw transactionError
```
Comment (alexfreska) [inline apps/explorer/app/contract/[id]/page.tsx:111]: <i class='graphite__hidden'>[Re: line +108]<br/><br/></i>

Can parallelize this at the top with other requests.<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/761?utm_source=unchanged-line-comment">Graphite</a>.</span>

apps/explorer/app/contract/[id]/page.tsx
Comment (telestrial) [reply apps/explorer/app/contract/[id]/page.tsx:42]: We didn't talk about this one yesterday. I just changed it to throw, but would you mind explaining your logic about why throw versus the other "successful" requests where we check for undefined and `return notFound()`? Trying to see if there's something here I need to know!

apps/explorer/components/Contract/index.tsx
telestrial: What about confirmation fields, for the same reason?

apps/explorer/app/contract/[id]/page.tsx
Comment (alexfreska) [reply apps/explorer/app/contract/[id]/page.tsx:42]: notFound/404 should be for when the user navigates to a contract ID that does not exist, so if we were to fetch the contract data and it 400s (not found) rather than 500s (server error). 
In this case, if the supporting rate data errors its definitely not a "contract not found", its a server error.

apps/explorer/components/Contract/index.tsx
alexfreska: I am not sure if we are able to fetch unconfirmed contracts, good question for the team.

apps/explorer/app/contract/[id]/page.tsx
Comment (telestrial) [reply apps/explorer/app/contract/[id]/page.tsx:42]: Okay, so does that mean that all of the return notFound's that are not specifically the contract call should be throws instead?

Comment (alexfreska) [reply apps/explorer/app/contract/[id]/page.tsx:42]: yeah I would first check the contract error, if its a 400 then return notFound, if the contract error is a 500 or any other required data fetching errors that can all be thrown to the error page.

apps/explorer/app/contract/[id]/page.tsx
Comment (telestrial) [reply apps/explorer/app/contract/[id]/page.tsx]: Leaving this open before I submit this PR because the issue is that we're not guaranteed confirmationIndex. It can be null.

apps/explorer/components/Contract/index.tsx
telestrial: Also leaving this open before I submit this PR: these are different but unlabeled. Not sure what the best course of action is on this one.

apps/explorer/app/contract/[id]/page.tsx
```diff
@@ -24,116 +24,115 @@ export function generateMetadata({ params }): Metadata {
...
+    to(explored.contractByID({ params: { id } })),
+    to<ExplorerFileContract[]>(explored.contractRevisions({ params: { id } })),
+    to(explored.consensusTip()),
   ])
 
-  if (error) {
-    throw error
-  }
+  if (rateError) throw rateError
+  if (!rate) throw 'No rate found in successful request'
```
Comment (alexfreska) [inline apps/explorer/app/contract/[id]/page.tsx:49]: if (!rate) throw new Error('No rate found in successful request')

apps/explorer/app/contract/[id]/page.tsx
```diff
@@ -24,116 +24,115 @@ export function generateMetadata({ params }): Metadata {
...
-  const contract = c?.contract
+  if (previousRevisionsError) throw previousRevisionsError
+  if (!previousRevisions)
+    throw 'No previousRevisions found in successful request'
 
-  if (!contract) {
-    return notFound()
-  }
+  if (currentTipError) throw currentTipError
+  if (!currentTip) throw 'No currentTip found in successful request'
```
Comment (alexfreska) [inline apps/explorer/app/contract/[id]/page.tsx:59]: if (!rate) throw new Error('No currentTip found in successful request')

apps/explorer/app/contract/[id]/page.tsx
```diff
@@ -24,116 +24,115 @@ export function generateMetadata({ params }): Metadata {
...
+    ChainIndex[]
+  >(
+    explored.transactionChainIndices({
+      params: { id: formationTransaction.id },
+    })
+  )
+
+  if (formationTxnChainIndicesError) throw formationTxnChainIndicesError
+  if (!formationTxnChainIndices)
+    return 'No formationTxnChainIndices found in successful request'
```
Comment (alexfreska) [inline apps/explorer/app/contract/[id]/page.tsx:110]: Is this supposed to be an error?

apps/explorer/app/contract/[id]/page.tsx
```diff
@@ -24,116 +24,115 @@ export function generateMetadata({ params }): Metadata {
...
 
   // Use the first chainIndex from the above call to get our parent block.
   const [parentBlock, parentBlockError] = await to(
-    explored.blockByID({ params: { id: transactionChainIndices[0].id } })
+    explored.blockByID({ params: { id: formationTxnChainIndices[0].id } })
   )
 
   if (parentBlockError) throw parentBlockError
-  if (!parentBlock) return notFound()
+  if (!parentBlock) return 'No parentBlock foudn in successful request'
```
Comment (alexfreska) [inline apps/explorer/app/contract/[id]/page.tsx:118]: `foudn`, and should this be an error?

apps/explorer/lib/contracts.ts
```diff
@@ -0,0 +1,21 @@
+import { ExplorerFileContract } from '@siafoundation/explored-types'
+
+export enum CONTRACT_STATUS {
```
alexfreska, line 3: I would rather not introduce enums into our codebase (https://www.youtube.com/watch?v=jjMbPt_H3RQ). Lets either do a `const` object or a union type.

apps/explorer/app/contract/[id]/page.tsx
Comment (alexfreska) [reply apps/explorer/app/contract/[id]/page.tsx:49]: I believe the existing code works even if rates error, so this should not throw.

apps/explorer/app/contract/[id]/page.tsx
```diff
@@ -24,116 +24,115 @@ export function generateMetadata({ params }): Metadata {
...
-  }
+  if (rateError) throw rateError
+  if (!rate) throw 'No rate found in successful request'
+
+  if (contractError) throw contractError
+  if (!contract) return notFound()
 
-  const contract = c?.contract
+  if (previousRevisionsError) throw previousRevisionsError
+  if (!previousRevisions)
```
Comment (alexfreska) [inline apps/explorer/app/contract/[id]/page.tsx:55]: A contract could have no previous revisions, this should not error.

alexfreska (commented):
> The exact timestamp for negotiation, expiration, proof deadline, and payout are a bit off from Sia Central

For all these slight differences, could attach a screenshot of a contract with the before and after, and we can have everyone double check that the differences are within reason.

> Do the resolved from/to buttons work as we would expect? I believe so but it's hard to find good test contracts that have both. I've only see one or the other. It shouldn't matter, but I like to visually/manually test as much as I can and couldn't find multiple contract renewals.

The following is a chain of 3 contracts: https://zen.siascan.com/contract/6344385f4189f14c78c8b793b93a8afebe2d138668becd8efc2057cb0d141557 that you can use to add an e2e test that validates the changes in this PR. Grab the key IDs and info of each from the production explorer and write a test that navigates through the renewals and verifies it matches up.


> Does the status make sense (in progress, obligation successful, obligation failed)? This is another one where I'd ideally find one of every kind but that's a little more detailed understanding than I may have right now.

Yes you should add a test for each at least success and failed, the above link includes a successful contract, lets have the team help find a failed example.

apps/explorer/app/contract/[id]/page.tsx
Comment (telestrial) [reply apps/explorer/app/contract/[id]/page.tsx:59]: What does this mean?

apps/explorer/components/Contract/index.tsx
alexfreska: Looks like throughout the PR we are swapping output IDs for addresses, lets ask whether we should add output ID to the API and continue to provide both, or change these lists to display these things as addresses only. The EntityListItem has a `type` we can set to address. Elsewhere such as on the Transaction page you have already migrated the EntityList renders OutputListItems which use and display both the output ID and the address (which navigates to the address page).
Screenshot 2024-10-25 at 9 00 44 AM
Screenshot 2024-10-25 at 9 00 09 AM

apps/explorer/app/contract/[id]/page.tsx
```diff
@@ -24,116 +24,115 @@ export function generateMetadata({ params }): Metadata {
...
     ),
   ])
 
-  const formationTransaction = ft?.transaction
-  const renewedFrom = formationTransaction?.contract_revisions?.[0]
-  const renewalTransaction = rt?.transaction
-  const renewedTo = renewalTransaction?.storage_contracts?.[0]
+  if (formationTransactionError) throw formationTransactionError
+  if (!formationTransaction)
+    throw new Error('No formation transaction found in successful reqeust')
```
Comment (n8maninger) [inline apps/explorer/app/contract/[id]/page.tsx:89]: `request`

apps/explorer/app/contract/[id]/page.tsx
```diff
@@ -24,116 +24,115 @@ export function generateMetadata({ params }): Metadata {
...
-    to(
-      explored.transactionChainIndices({
-        params: { id: formationTransaction?.id || '' },
-      })
-    ),
-    to(explored.consensusTip()),
-  ])
+  if (renewalTransactionError) throw renewalTransactionError
+  if (!renewalTransaction)
+    throw new Error('No renewal transaction found in successful reqeust')
```
Comment (n8maninger) [inline apps/explorer/app/contract/[id]/page.tsx:93]: `request`

n8maninger (changes requested):
Couple of small UX thoughts, couple of spelling nits, but overall good work.

apps/explorer/app/contract/[id]/page.tsx
Comment (alexfreska) [reply apps/explorer/app/contract/[id]/page.tsx:89]: still says `reqeust`

apps/explorer/app/contract/[id]/page.tsx
Comment (alexfreska) [reply apps/explorer/app/contract/[id]/page.tsx:93]: still says reqeust

apps/explorer/components/Contract/index.tsx
```diff
@@ -40,114 +54,109 @@ export function Contract({ contract, rates, renewedFrom, renewedTo }: Props) {
...
-        value: String(contract.proof_confirmed),
+        value: String(!!contract.proofTransactionID),
       },
       {
         label: 'negotiation height',
         copyable: false,
-        value: contract.negotiation_height?.toLocaleString() || '-',
+        value:
+          (
+            contract.confirmationIndex?.height ||
```
alexfreska, line 77: In what scenario does this fallback to the formation height? Might be a good comment to add.

apps/explorer/components/Contract/index.tsx
```diff
@@ -274,39 +285,40 @@ export function Contract({ contract, rates, renewedFrom, renewedTo }: Props) {
...
     return false
   }
 
   const { renterPayoutValid, renterPayoutMissed } =
     getNewContractFormattedOutputs(contract)
 
   // renter payout valid and missed should be the same
-  if (renterPayoutValid.value !== renterPayoutMissed.value) {
+  if (renterPayoutValid?.value !== renterPayoutMissed?.value) {
+    // Does we need to catch undefined cases now?
```
alexfreska, line 303: `Do* we need`

apps/explorer/components/Contract/index.tsx
telestrial: I believe the answer is: between when a contract is formed and when it is confirmed but I'm not positive. I know on the frontend side that we're not guaranteed the confirmationIndex. Does that answer track for you?

n8maninger: The contracts route should always return a confirmation index. Probably good to keep this check, but we don't index contracts that have not been confirmed.

telestrial: This is what I'm responding to, in terms of Go. The `/contracts` route returns `ExtendedFileContract`, which has this:

```go
ConfirmationIndex         *types.ChainIndex    `json:"confirmationIndex"`
ConfirmationTransactionID *types.TransactionID `json:"confirmationTransactionID"`
```

alexfreska: Are you referring to the fact that its a pointer? Although a pointer _could_ be null, the business logic may still be designed to realistically always return a value. @n8maninger is there a way we could make it more clear whether or not the json would actually return a null value?

n8maninger: Unclear why it's a pointer for me as well. Only confirmed contracts are indexed, so it should never be nil.

telestrial: I think "could be null" is a good reason to expect in the code that it might be null. Otherwise, it feels to me like we would be hiding an understanding between server and client from consumers. It's quite possible that it's never an issue. But, if the typings match and something goes wrong, it's likely going to be easier to diagnose than if they don't match--from a consumer's perspective. As with anything I write, I'm always happy to defer to greater wisdom here but my view is that it should stay possibly null unless the server-side typing say otherwise.

That said, is `|| formationTxnChainIndex[0].height` the move here or should we display "-" in lieu of the `confirmationIndex`? Looking at this more, we really never reach the "-" case here. We're either getting that formationTxChainIndex or we're error'ing this page, so I could easily delete the OR case here and it would actually be a bit cleaner.

alexfreska: Since we know the field will never be null (because only confirmed contracts are indexed), I would lean towards typing it on our end as always defined. If we could also change the type on the Go side to make this fact more clear, that would even better.

@n8maninger is it accurate for us to assume pointers are used because they can be null, and the field would otherwise not be a pointer? Or are they sometimes used in places where they will never be null for other reasons?

telestrial: > Every other instance with pointers I can think of right now is somewhere that can be null and should be explicitly handled.

@n8maninger Does that include confirmationTransactionID?

n8maninger: Both `confirmationTransactionID` and `confirmationIndex` are no longer pointers in latest

telestrial:
Got this covered now in a down stream PR.

apps/explorer/components/Contract/index.tsx
telestrial: Got this covered now in a downstream PR.

n8maninger (approved):

PR #799, opened 2024-10-29
fix(explorer): fix network request calls when navigating to block by height

telestrial:

PR #800, opened 2024-10-29
test(explorer): add e2e tests for all routes

telestrial:
This all feels pretty straight-ahead to me except for my approach to the hosts route. Here, I chose to use regex to test the format of the different displayed values. If the value matches one of any pattern (found in `constants.ts`, the tests past. I weighed this option against the following:

* This route is server-side, making mocking tough. I think we'd have to pick up a dependency to do something like that here, and I didn't feel strongly enough about it for this one case.
* Snapshot testing didn't feel like the correct tool for this job and maintaining it seemed like a bad idea considering how quickly host settings can change.
* Testing at a lesser level i.e. "is there a value here?" is not really a test worth writing, imo. The props will help check this. Using the values without them being there will also help us.
* Similar to our historical data checking (looking up contracts we know won't move any more and relying on those values, etc), with the regex approach, if a format changes for hosts, the tests will break. This could be intentional or not intentional, and then we look for bugs or change the test, depending. That feels like the same type of pattern we're using elsewhere...just a bit harder to read, on its face.

I'm totally open to other ideas here, but that's why I made that particular decision.


telestrial:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/800?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#800** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/800?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#799** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/799?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

apps/explorer-e2e/src/fixtures/constants.ts
```diff
@@ -0,0 +1,92 @@
...
+export const APP_NAME = 'explorer-identity-appName'
+export const APP_NETWORK = 'explorer-identity-network'
+export const DATUM_VALUE = 'explorer-datum-value'
+
+// Home page
+export const METRICS_ITEM = 'explorer-metrics-item'
+export const LATEST_BLOCKS_ITEM = 'explorer-latestBlocks-item'
+export const TOP_HOSTS_ITEMS = 'explorer-topHosts-item'
+
+// Contract page
```
alexfreska, line 12: I would lean towards colocating these strings with actual fixtures / helpers they are used in rather than the extra indirection of a generic constants file, for example SEARCHBAR is only used in ExplorerApp it could at least be in that file or inlined. The renew button strings could be in a contracts fixture file or inlined there in methods like `clickRenewedFrom`.

No need to change now, just some thoughts to consider in future iteration.

apps/explorer-e2e/src/fixtures/constants.ts
```diff
@@ -0,0 +1,92 @@
...
+// Host page
+export const HOST_PRICING = 'explorer-hostPricing'
+export const HOST_SETTINGS_PATTERNS = [
+  '\\b\\d+(?:\\.\\d{1,2})?\\s?(TB|GB|MB|KB|B)\\b', // 3.00 TB
+  '(\\$\\d+(?:\\.\\d{1,3})?|\\d+(?:\\.\\d{1,3})?\\s?[A-Z]{2})\\/(TB|GB|MB|KB|B)\\/(month|year|day)', // $1.16/TB/month
+  '(\\$\\d+(?:\\.\\d{1,3})?|\\d+(?:\\.\\d{1,3})?\\s?[A-Z]{2})\\/(TB|GB|MB|KB|B)', // $1.12/TB
+  '\\b\\d+\\s?(days?|months?|years?)\\b', // 6 months
+  '^\\$\\d+(?:\\.\\d+)?$', // $4.50
+  '(\\$\\d+(?:\\.\\d+)?|\\d+(?:\\.\\d+)?\\s?[A-Z]{2})\\/million', // $0.0045 SC/million
+  '^\\d+(?:\\.\\d+)?\\s?(KS|mS)$', // 1.000 KS/mS
```
alexfreska, line 25: I am not sure the value of these regex settings and price table tests as they will essentially match anything besides a string, it might provide a little more signal to grab a specific field and test against a specific one of these, but that also feels pretty redundant. Lets stick with this for now and see if we feel it provides some stability.

libs/design-system/src/components/EntityAvatar.tsx
```diff
@@ -34,7 +34,7 @@ export function EntityAvatar({
     />
   )
   const linkEl = href && (
-    <Link href={href} underline="none">
+    <Link href={href} underline="none" data-testid={`entity-link-${href}`}>
```
alexfreska, line 37: This seems a bit redundant, we can give it a go but I would probably have tried to leverage the existing href in combination with a general testid. Combining more widely useful testids means we don't have to put a testid on specific elements for every test case.
```
page.getByTestId('entity-avatar').locator(`a[href*=${href}]`)
```

alexfreska:
> This route is server-side, making mocking tough. I think we'd have to pick up a dependency to do something like that here, and I didn't feel strongly enough about it for this one case.
> Snapshot testing didn't feel like the correct tool for this job and maintaining it seemed like a bad idea considering how quickly host settings can change.
Yeah, it may be good to figure something out here, but I think seeing how far we get without it first is a good approach. Also, I think adding some unit tests with a variety of example states/data for the host and contract page will be a bit easier to setup and would give us some of this. 
I'm glad you avoided snapshot tests for now, I do like them in certain senarios but I think its best to first add non-snapshot tests and then sprinkle them in strategically, going all in too fast can sometimes cause some testing snowblindness.

> Testing at a lesser level i.e. "is there a value here?" is not really a test worth writing, imo. The props will help check this. Using the values without them being there will also help us.
> Similar to our historical data checking (looking up contracts we know won't move any more and relying on those values, etc), with the regex approach, if a format changes for hosts, the tests will break. This could be intentional or not intentional, and then we look for bugs or change the test, depending. That feels like the same type of pattern we're using elsewhere...just a bit harder to read, on its face.
Yeah this makes sense, as I commented these may not provide as much value as a full test with stable data, but lets give it a go and see if it provides some stability.

alexfreska (approved):
Great stuff! Left a few comments to think about, but lgtm.

libs/design-system/src/components/EntityAvatar.tsx
telestrial: I went ahead and adjusted this a bit. Instead of such a specific ref, I used "entity-link"--I see value in having these links instrumented, but I adjusted the selector in a similar way to what you suggested:

```ts
await page
    .locator(
      'a[data-testid="entity-link"][href*="8a03682f22857f306a95f55a28fa9edf"]'
    )
    .click()
```

telestrial:
### Merge activity

* **Nov 5, 2:17 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/SiaFoundation/web/800).

PR #843, opened 2024-12-06
fix(website): rpm ARM64 download link named AMD64

sergeypdev:


alexfreska (approved):
@sergeypdev good catch and thank you for putting in the fix!

PR #844, opened 2024-12-10
refactor: improve multiselect behaviours

alexfreska:
### changes
- Multi-select now supports single select, toggle select, and range selection interactions, with click, ctrl-click, and shift-click.
- Table now prevents default on any mouse down with shift held. This ensures the user does not highlight text while shift selecting a range of rows.
- Renamed selectionMap to selection in the useMultiSelect hook API.

### overview
**normal click (single select):**

- The selection is reduced to the clicked row, and the range selection anchor is updated.
- Clicking the only selected row does not deselect it.

**ctrl/cmd-click (toggle select):**

- Toggles selection of an individual row without affecting others.
- If selected, updates the anchor to that row.
- If deselected, the anchor is updated to the nearest selected index, first checking higher indices, then lower. If no index is found reset the anchor.

**shift-click (range select):**

- Always select the full continuous range between the anchor and the clicked row, deselect all rows between the anchor and the last range end from the same anchor.
- Do not update the anchor on shift-click, unless there is none set yet.
- If no anchor is set, the range select will use the top row as the anchor and set the range end to the selected row.

**anchor**

- Whenever the anchor is updated, the range end is reset.
- Whenever the anchor is reset the range end is also reset.

### video
I could not get screenstudio to show the keypresses, but the following video shows some of the new multi-select interactions, such as how ranges are selected and rows between the anchor and the previous range end are deselected.

https://github.com/user-attachments/assets/4c49769c-2fad-4a27-8063-48090637fa5f


alexfreska:
* **#844** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/844?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/844?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

peterjan (approved):
sweet

PR #845, opened 2024-12-10
feat(design-system): create ReactHookFormField components file

telestrial:
The purpose of this PR is to create a component file comparable to `FormFormik.ts` to carry the `formik` to `react-hook-form` transition--the ultimate goal being to delete `FormFormik` and the `formik` dependency across the project. It may be helpful to pull that file up while reviewing. While I was doing this translation, I took the opportunity to refactor a bit:

1. I moved the individual component declarations for `TextField`, `NumberField`, and `SiacoinField` into the overall FormField component, reducing the abstractions here by one. I saw the purpose of those abstractions to center around setting defaults, so:
2. I moved the defaults that those individual abstractions were setting into the unwrapped components' props. Example: `<Numberfield...allowDecimals={allowDecimals || false}`. All these props were already flowing through the parent component, so I decided to use the base components sooner and trim down the complexity.

There is a question/potential issue to rectify that may impact this PR, but I'll talk about it on top of this stack of two as the issue is passed down. It has to do with the typing of the props/genericism of the `ReactHookFormField` component with `react-hook-form`'s `FieldValues`.

PR #846, opened 2024-12-10
refactor(explorer): move faucet form from formik to react-hook-form

telestrial:
This replaces formik in the explorer faucet fund form. This doesn't remove the `formik` dependency, yet, because we still have these files using formik:

```
ContractsFilterContractIdDialog
ContractsFilterAddressDialog
ContractsFilterPublicKeyDialog
HostsFilterAddressDialog
HostsFilterPublicKeyDialog
AllowlistForm
BlocklistForm
FileCreateDirectoryDialog
WalletAddAddressDialog
```

I have one concern that I really haven't found the right solution for. I think a blue sky implementation of react-hook-form looks something like this:

```react
const { handleSubmit, register } = useForm<(<foo: string, bar: number)>()
const onSubmit = (data) => { data.foo... }

return <form onSubmit={handleSubmit(onSubmit)}>
         <input type="text" {...register('foo')} />
       </form>
```

In theory, this should extend react-hook-form's `FieldValues` to be properly typed and, for example, prevent trying to register a property that doesn't exist in the type or use a value of the wrong type. The problem? I can't figure out how to receive that type in the ReactHookForm component. I've tried pretty much everything I can think of, including a `T Extends FieldValues` approach, but it hasn't worked. In lieu of that panning out, this implementation does work-setting the type in the data of the `onSubmit`. But you won't be properly checked around registering and setting values the way you might be if we could receive the type correctly in the ReactHookFormField component's `FieldValues`. Open to ideas here, for sure.

Otherwise, this does remove `formik` for `react-hook-form` in the faucet form. There just may be a better way to type it on the component side which would somewhat affect this implementation.

### DIFF
--- a/.changeset/brave-rules-beam.md
+++ b/.changeset/brave-rules-beam.md
@@ -1,0 +1,5 @@
+---
+'explorer': minor
+---
+
+Refactored faucet form from formik to react-hook-form.
--- a/apps/explorer/components/Faucet/FaucetFundForm.tsx
+++ b/apps/explorer/components/Faucet/FaucetFundForm.tsx
@@ -1,94 +1,109 @@
 import {
-  FormFieldFormik,
-  FormSubmitButtonFormik,
   Paragraph,
+  ReactHookFormField,
+  ReactHookFormSubmitButton,
   triggerSuccessToast,
+  triggerErrorToast,
 } from '@siafoundation/design-system'
 import { toHastings } from '@siafoundation/units'
 import BigNumber from 'bignumber.js'
-import { useFormik } from 'formik'
-import * as Yup from 'yup'
 import { networkName } from '../../config'
 import { useFaucetFund } from '../../hooks/useFaucetFund'
+import { useForm } from 'react-hook-form'
 
-const initialValues = {
-  address: '',
-  amount: undefined as BigNumber | undefined,
+type FaucetFundFormFields = {
+  address: string
+  siacoin: BigNumber
 }
 
-const validationSchema = Yup.object().shape({
-  address: Yup.string().required('Required'),
-  amount: Yup.string()
-    .required('Required')
-    .test(
-      'greater than zero',
-      'Must be greater than 0 SC',
-      (val) => !new BigNumber(val || 0).isZero()
-    )
-    .test('less than 1000', 'Must be 50,000 SC or less', (val) =>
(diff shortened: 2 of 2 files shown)


telestrial:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/846?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#846** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/846?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/846?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#845** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/845?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

alexfreska:
@telestrial check out the `form/` folder in the design-system it contains a whole bunch of useful components for using `react-hook-form`. If you check out any of the existing forms (eg: https://github.com/SiaFoundation/web/blob/main/apps/walletd/dialogs/WalletUpdateDialog/index.tsx) you can see they all follow the same pattern where they define a `fields` or `getFields` which provides field configuration data of type `ConfigFields`, all the components use this data format, which streamlines things. They all also define a `defaultValues` or `getDefaultValues` from which we get a `Values` type and that we pass to useForm to give it initial data and type information. If you look at a few of these existing forms it should give you some good information.

alexfreska (commented):
@telestrial check out the `form/` folder in the design-system it contains a whole bunch of useful components for using `react-hook-form`. If you check out any of the existing forms (eg: https://github.com/SiaFoundation/web/blob/main/apps/walletd/dialogs/WalletUpdateDialog/index.tsx) you can see they all follow the same pattern where they define a `fields` or `getFields` which provides field configuration data of type `ConfigFields`, all the components use this data format, which streamlines things. They all also define a `defaultValues` or `getDefaultValues` from which we get a `Values` type and that we pass to useForm to give it initial data and type information. If you look at a few of these existing forms it should give you some good information.

PR #852, opened 2024-12-12
refactor(renterd): min account expiry and min price table validity units

alexfreska:
- The min account expiry and min price table validity configuration settings are now stored in milliseconds.

PR #853, opened 2024-12-12
fix(renterd): first file download no bucket error

alexfreska:
- Fixed an issue where the first attempt to download a file would show a bucket not found error.

### Notes
- Strange issue where buckets.data is still undefined in the closure even though the data has been fetched. Adding useCallback does the trick, oddly even just console logging buckets.data above downloadFiles fixes the issue.

### DIFF
--- a/.changeset/late-points-learn.md
+++ b/.changeset/late-points-learn.md
@@ -1,0 +1,5 @@
+---
+'renterd': patch
+---
+
+Fixed an issue where the first attempt to download a file would show a bucket not found error.
--- a/apps/renterd-e2e/src/specs/files.spec.ts
+++ b/apps/renterd-e2e/src/specs/files.spec.ts
@@ -111,6 +111,29 @@ test('can upload, rename, and delete files', async ({ page }) => {
   await deleteBucket(page, bucketName)
 })
 
+test('can upload and download a file', async ({ page }) => {
+  const bucketName = 'files-test'
+  const fileName = 'sample.txt'
+  const filePath = `${bucketName}/${fileName}`
+
+  // Create bucket.
+  await navigateToBuckets({ page })
+  await createBucket(page, bucketName)
+  await openBucket(page, bucketName)
+
+  // Upload.
+  await dragAndDropFileFromSystem(page, fileName)
+  await expect(page.getByText('100%')).toBeVisible()
+  await fileInList(page, filePath)
+
+  // Download.
+  await openFileContextMenu(page, filePath)
+  await page.getByRole('menuitem', { name: 'Download file' }).click()
+  await expect(
+    page.getByRole('button', { name: 'Active downloads' })
+  ).toBeVisible()
+})
+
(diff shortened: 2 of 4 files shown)


alexfreska:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/853?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#853** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/853?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/853?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#852** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/852?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

telestrial (approved):
> Strange issue where buckets.data is still undefined in the closure even though the data has been fetched. Adding useCallback does the trick, oddly even just console logging buckets.data above downloadFiles fixes the issue.

I've been mulling this over a bit because it is an interesting issue! On its face, it could appear to treat a symptom and not the cause. That's what I initially thought, but after thinking a bit, here's what I believe is going on:

When you define a function like `downloadFiles`, it captures the state of `buckets.data` at that moment. React doesn’t create a new function reference on subsequent renders unless explicitly instructed to (as with `useCallback`), which is why you're seeing a stale reference. The closure retains the value of `buckets.data` as it was when the function was first defined, without updating automatically.

The reason `console.log` seems to work is that it forces the runtime to evaluate the current value of `buckets.data` at the time the log statement is executed. So console.log() is the current render and `downloadFiles` is likely the first render.

All that said, I believe this is exactly the right answer to the problem. It's probably more of a use case for `useCallback` than most other common uses.

peterjan (approved):

ChrisSchinnerl (approved):

alexfreska:
@telestrial it has to be something with a stale closure but when `downloadFiles` is not wrapped in `useCallback` React does create a new function reference **every** re-render, for example if I were to try to use it as a dependency for another hook deps you would get a warning like:
```
'downloadFiles' function makes the dependencies of useEffect Hook (at line 153) change on every render. To fix this, wrap the definition of 'downloadFiles' in its own useCallback() Hook.
```
The behaviour you describe should be when you do a useCallback with empty deps `[]`. But I am really not sure - definitely an interesting Advent of Bugs problem haha.

alexfreska:
Single test failing due to: https://github.com/SiaFoundation/renterd/issues/1747

alexfreska:
### Merge activity

* **Dec 13, 11:43 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/SiaFoundation/web/853).

PR #897, opened 2025-01-21
feat(explorer): use explored for search

telestrial:
This PR replaces Sia Central with explored, for the explorer search feature.

Down stack, we add the 'host' SearchResultType return. Before the [connected](https://github.com/SiaFoundation/explored/pull/161) PR hits, if a user searches for a host pubkey with the `ed25519` prefix, we will succeed. If not prefixed, we will hit the error case (and tell the user the request failed) but, after the PR hits, this should just work.

### DIFF
--- a/apps/explorer/components/Layout/Search.tsx
+++ b/apps/explorer/components/Layout/Search.tsx
@@ -11,9 +11,9 @@ import { Search16 } from '@siafoundation/react-icons'
 import React, { useCallback } from 'react'
 import { useRouter } from 'next/navigation'
 import { useForm } from 'react-hook-form'
-import { useSiaCentralSearch } from '@siafoundation/sia-central-react'
 import { routes } from '../../config/routes'
-import { siaCentralApi } from '../../config'
+import { explored } from '../../config/explored'
+import { to } from '@siafoundation/request'
 
 const defaultValues = {
   query: '',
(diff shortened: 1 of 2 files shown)


telestrial:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/897?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#897** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/897?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/897?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#896** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/896?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

apps/explorer/components/Layout/Search.tsx
```diff
@@ -37,64 +37,73 @@ export function Search() {
...
+        case 'block':
+          router.push(routes.block.view.replace(':id', values.query))
+          break
+        case 'contract':
+          router.push(routes.contract.view.replace(':id', values.query))
+          break
+        case 'host':
+          router.push(routes.host.view.replace(':id', values.query))
+          break
+        case 'siacoinElement':
```
alexfreska, line 83: Why do these go to contract?

alexfreska (approved):

apps/explorer/components/Layout/Search.tsx
telestrial: I'm following the `explored` implementation here but I'll be the first to say that I don't fully understand the use case:

`SearchTypeSiacoinElement means we found a contract with the given ID.`

So my assumption, based on the comment, is that this route direct will work but it may not be a common way of using the information from an output.

@chris124567: What is the intended use case behind `SearchTypeSiacoinElement` and `SearchTypeSiafundElement`? What would a user be trying to do here and with what piece of information?

n8maninger: `SearchTypeSiacoinElement` and `SearchTypeSiafundElement` point to UTXOs. The comment is copy pasta.

telestrial: Okay. Then, I'll delete these two cases and have this hit the default `unsupported` case for now--until we have distinct UI for UTXOs.

chris124567: I'll fix those comments :)

PR #903, opened 2025-01-24
fix(renterd): files name sort is on path

alexfreska:
- Fixed an issue where files were being sorted by name rather than full path in all files mode.
  - The sort field is called `name` so it was sorting on the `name` on the client - the server is really sorting on the full path when returning delimiter="" results.
  - The reason we have to re-apply the sorting at all on the client-side because we naturally weave in any active uploads that fit into the current directory or the all files view. This is to achieve a UX where uploads show immediately just in a "syncing" state as seen in iCloud, Dropbox, etc.
- The default files sort direction is now asc instead of desc.

### DIFF
--- a/.changeset/healthy-moles-unite.md
+++ b/.changeset/healthy-moles-unite.md
@@ -1,0 +1,5 @@
+---
+'renterd': patch
+---
+
+Fixed an issue where files were being sorted by name rather than full path in all files mode.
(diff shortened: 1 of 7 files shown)


alexfreska:
* **#903** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/903?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/903?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

telestrial (approved):

PR #900, opened 2025-01-22
feat(explored-types): add exchange rate API

telestrial:

PR #907, opened 2025-01-28
feat(explorer): use explored exchange rates

telestrial:


telestrial:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/907?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#907** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/907?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/907?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#900** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/900?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

apps/explorer/components/SCToFiatDisplay.tsx
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export function SCToFiatDisplay({ sc }: Props) {
+  const {
+    settings: { currency },
+  } = useAppSettings()
+
+  const { rate, error, isLoading } = useSiascanExchangeRate({
+    currency: currency.id,
+    api: 'https://api.beta.siascan.com/zen/api',
```
alexfreska, line 22: Since this is temporary lets make a ticket to remove it close to launch.

alexfreska, line 22: Also maybe just  remove `zen`, not sure the difference but might as well use `mainnet` exchange rates.

apps/explorer/components/SCToFiatDisplay.tsx
```diff
@@ -0,0 +1,33 @@
...
+  })
+
+  const currencyOption = currencyOptions.filter(
+    (option) => option.id === currency.id
+  )[0]
+
+  if (isLoading) return <Skeleton className="h=[28px] w-[54px]" />
+  if (error) return ''
```
alexfreska, line 30: Should we display a warning icon and a tooltip maybe? or at least a dash with a tooltip.

alexfreska (commented):
Looking good, few bits of feedback. Also re-run the test suite and if you still see any errors instantiating the cluster please report this in our engineering channel so the core team can do this fix.

telestrial:
The component I added works for many cases but there are still some where we only need the value. A place where we use `getDownloadCost` or something akin. I may end up reworking the component and decoupling the fetching from the displaying. Still thinking through the best pattern here.

Will do, on the cluster note.

apps/explorer/components/SCToFiatDisplay.tsx
telestrial: The dash and tooltip seems right but I'll look at this. I think a warning type of icon with a tooltip could work, too, but I'd worry about multiple instances of it on one page. I'm not sure if that happens anywhere. Will investigate.

telestrial:
While I know this is a draft, I'm going to start over. There are assumptions in this PR about the ways exchange rates are used that are incomplete. Instead of backing down the path, trying to account for those, and changing my solution to match, I'd rather start from 0 with those in mind. I may even ultimately end up in a similar place, but I'd rather have the full picture first and then solve.

PR #919, opened 2025-02-10
feat(units): add remainingStorage function for ExplorerHost

telestrial:

PR #918, opened 2025-02-10
feat(explored-types): add hostsList API

telestrial:

PR #915, opened 2025-02-05
feat(explorer): remove host benchmarks

telestrial:
Slight visual changes on this one for the hosts list on the home page to fix some consequences of removing the two benchmark subvalues:

* Remaining storage moves to a "primary" value with its own icon
* Removed justify-between for these now four items - this felt too spaced out with the smaller amount of text. Runner up visually for me was justify-around, but I like unset better.
* ~~At the lower width breakpoint, now storage cost and download cost stay visible.~~ Downstack, we've introduced 2xs and 3xs container breakpoints to `theme` and used container queries to makes this more responsive.

PR #920, opened 2025-02-10
feat(explorer): use explored for top hosts call/ui

telestrial:


telestrial:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/920?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#920** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/920?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/920?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#919** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/919?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#918** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/918?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#915** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/915?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#916** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/916?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#917** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/917?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

apps/explorer/app/page.tsx
```diff
@@ -35,30 +35,43 @@ export default async function HomePage() {
       })
     ),
     to(
-      siaCentral.hosts({
-        params: {
-          limit: 5,
-        },
+      explored.hostsList({
+        params: { sortBy: 'date_created', dir: 'asc' },
+        data: { Online: true },
```
alexfreska, line 40: this is capitalized?

apps/explorer/app/page.tsx
```diff
@@ -35,30 +35,43 @@ export default async function HomePage() {
...
       })
     ),
     to(explored.hostMetrics()),
     to(explored.blockMetrics()),
   ])
 
+  const sortedTopHosts = topHosts
```
alexfreska, line 47: If we are filtering/sorting on the client, I am assuming the endpoint return **all** matching hosts without a limit?

telestrial: We can place a limit on the request, but I chose otherwise to improve the likelihood that the "top 5" hosts would be exemplary. This request happens server-side--which I think helps. I could be overthinking it, though. The issue I was wrestling with was sample size. 

You either take the X oldest hosts or you take the X highest uptime hosts to start. If you just take the top 5 oldest hosts, say, they could have inferior uptime. If you take the top 5 uptime, you would definitely get hosts that have not proven longevity.

So I think you have to do something on the client side because you are doing two things here--age and uptime--but also because you need to take more hosts than you plan to display.

How many more should you take to get an exemplary result? It's subjective/indeterminate.

alexfreska: Yeah we need the full dataset to provide accurate results - if we just sort and limit on a single field like age there is a good chance we are clipping out the "top" hosts. Ideally the whole ranking would be done in the database, if not, we should pull down the whole dataset. In the future this may become prohibitive but right now the active hosts number is only around 500 so we can just pull them all in, and just make sure to cache the results for a few minutes to prevent excessive queries.
In terms of the actual ranking algorithm what if we did a weighted score across age, pricing, used storage, uptime, so we factor in more of the signals a user would consider when deciding if a host is good. Heres a rough gist of what I was thinking could work:

```js
function normalize(value, min, max, invert = false) {
  if (max === min) return 0
  const normalized = (value - min) / (max - min)
  return invert ? 1 - normalized : normalized
}

function rankHosts(hosts, weights) {
  const minMax = (key) => ({
    min: Math.min(...hosts.map((h) => h[key])),
    max: Math.max(...hosts.map((h) => h[key])),
  })

  const ranges = {
    age: minMax('age'),
    uptime: { min: 0, max: 100 }, // Assuming uptime is in percentage
    downloadPrice: minMax('downloadPrice'),
    uploadPrice: minMax('uploadPrice'),
    storagePrice: minMax('storagePrice'),
    usedStorage: minMax('usedStorage'),
  }

  return hosts.map((host) => ({
    host,
    score:
      weights.age * normalize(host.age, ranges.age.min, ranges.age.max) +
      weights.uptime * normalize(host.uptime, ranges.uptime.min, ranges.uptime.max) +
      weights.downloadPrice * normalize(host.downloadPrice, ranges.downloadPrice.min, ranges.downloadPrice.max, true) +
      weights.uploadPrice * normalize(host.uploadPrice, ranges.uploadPrice.min, ranges.uploadPrice.max, true) +
      weights.storagePrice * normalize(host.storagePrice, ranges.storagePrice.min, ranges.storagePrice.max, true) +
      weights.usedStorage * normalize(host.usedStorage, ranges.usedStorage.min, ranges.usedStorage.max),
  })).sort((a, b) => b.score - a.score) // Sort descending
}

// The above assumes data looks something like the following, normalizing each data point to a value between 0 and 1:
const hosts = [
  { age: 5, uptime: 99.9, downloadPrice: 10, uploadPrice: 5, storagePrice: 2, usedStorage: 500 },
  { age: 3, uptime: 99.5, downloadPrice: 15, uploadPrice: 7, storagePrice: 1.5, usedStorage: 300 },
  { age: 7, uptime: 99.8, downloadPrice: 8, uploadPrice: 4, storagePrice: 2.5, usedStorage: 700 },
]

// The weights can be adjusted to boost whatever factor we want to consider the the most.
const weights = {
  age: 0.3,
  uptime: 0.3,
  downloadPrice: 0.1,
  uploadPrice: 0.1,
  storagePrice: 0.1,
  usedStorage: 0.1,
}
```

apps/explorer/app/page.tsx
```diff
@@ -35,30 +36,33 @@ export default async function HomePage() {
...
+      explored.hostsList({
+        params: { sortBy: 'date_created', dir: 'asc', limit: 500 },
+        data: { online: true },
       })
     ),
     to(explored.hostMetrics()),
     to(explored.blockMetrics()),
   ])
 
+  const selectedTopHosts = rankHosts(topHosts)
```
alexfreska, line 48: This request and transformation of host data doesn't need to and probably shouldn't re-rerun on every page request, could you cache it? I think you can do this by wrapping it with `unstable_cache` and adding a revalidate time, but definitely double check the nextjs 14 docs.

alexfreska (approved):

telestrial:
### Merge activity

* **Feb 13, 2:57 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/SiaFoundation/web/920).

PR #933, opened 2025-02-18
feat(explorer) move host opengraph lookup to explored

telestrial:

PR #935, opened 2025-02-18
feat(explorer): move exchange rate lookups to explored

telestrial:
This is a bigger one! This replaces all Sia Central exchange rate API calls to explored API calls for `explorer`. One medium-sized detail that was introduced here:

* Visually, I decided to use an `animate-pulse` loading state for the different currency-based values (`$0.00/TB`, `$0.00/TB/Month`, etc). I did try out existing components, such as `LoadingDots` or `Skeleton`, but these components do not preserve the layout. With this change, all the affected pages move to or very close to `0.00` cumulative layout shift. It also has the advantage of not showing Siacoins then replacing to the client's chosen currency. Less movement from the initial paint, overall, across `Home`, `Contracts`, and `Host`.
* However, a small wrinkle is that we'll now show the `$` symbol for a moment even if we are set to a different currency. In production on average connections, I suspect this is a very very small amount of time, if even perceptible. I am open to workarounds for this, but, in my testing (and I could have just misunderstood the proper way to use this or I'm unaware of other approaches), `useActiveSiacscanExchangeRate` only supplies that prefix once the rate has been retrieved. Therefore, we can't use that ahead of time because, by the time we have it, we're ready to display the loaded value. If we have another way to get this, I would pursue integrating that into the `LoadingCurrency` component so that we show `<client's chosen currency symbol>0.00/whatever` while loading. That would be ideal.


telestrial:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/935?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#936** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/936?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#935** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/935?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/935?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#934** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/934?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#933** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/933?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

alexfreska (commented):
There is a way to server render the initial value here which would allow us to avoid the initial loading states entirely. We can fetch the exchange rate in the root layout server component and then pass it as fallback data to the `swr` provider for the specific `swr` key. This allows the exchange rate hook which fetches on the client-side to start with the user specific currency as initial data.
- This required some plumbing changes, I added support here: https://app.graphite.dev/github/pr/SiaFoundation/web/939/feat-app-router-swr-fallback-data
- And this is how to use the added feature: https://app.graphite.dev/github/pr/SiaFoundation/web/940/feat-explorer-server-side-currency-data
- If all looks good, I can move #939 below all these changes and you can squash #940 into your PR.

Besides the above, some notes on the PR:
- Regarding loading states, seeing a specific currency $ and 0 values is a bit more confusing then seeing a skeleton since the user is essentially seeing a specific value - this is similar to seeing the wrong date and then the correct user date.
- Removing the exchange rate fetching per page is good, and actually fits with doing it only in the root layout.
- The loading state are probably still necessary for when the user switches the currency on the client-side.

apps/explorer/components/Host/HostSettings.tsx
```diff
@@ -15,17 +14,19 @@ import {
...
 type Props = {
   host: ExplorerHost
-  rates?: SiaCentralExchangeRates
 }
 
-export function HostSettings({ host, rates }: Props) {
-  const exchange = useExchangeRate(rates)
+export function HostSettings({ host }: Props) {
+  const exchange = useActiveCurrencySiascanExchangeRate({
+    api: 'https://api.beta.siascan.com/api',
```
alexfreska, line 28: We should move this string out to the config file so its easy to swap all instances when we release. Also, if we include the ssr changes we can use the variable for the fallback key as well.

alexfreska (approved):

telestrial:
### Merge activity

* **Feb 21, 1:58 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/SiaFoundation/web/935).

PR #937, opened 2025-02-18
refactor(explorer): use client side rendering for timestamps

telestrial:
This PR puts the result of `blockHeightToHumanDate` operations within the `explorer` totally on the client's side, using a ... `animate-pulse` fallback. As far as client is concerned, I expect them to barely experience this fallback given the speed of the operation (EDIT: looking at the live preview, this appears to be true. It's basically imperceptible.). We do also benefit visually here by not showing the server timestamp and then switching to the client's. However, the biggest gain is in local development under `strict mode`, as affected components were switching to client-side rendering. This should help us more accurately assess clients' loading experience in development, which I believe is a strong reason to take this approach.

### DIFF
--- a/apps/explorer/components/Transaction/TransactionHeader.tsx
+++ b/apps/explorer/components/Transaction/TransactionHeader.tsx
@@ -2,6 +2,7 @@ import { Badge, Text, Tooltip } from '@siafoundation/design-system'
 import { humanDate } from '@siafoundation/units'
 import { routes } from '../../config/routes'
 import { EntityHeading } from '../EntityHeading'
+import LoadingTimestamp from '../LoadingTimestamp'
 
 export type TransactionHeaderData = {
   id: string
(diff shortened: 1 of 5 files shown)


alexfreska:
Whats the benefit here over just `suppressHydrationWarning={true}` on these elements? I guess that would that render a server localized datetime, and them immediately update to the user localized just like this does right? In that case, maybe skeleton=>user datetime is better than server datetime=>user datetime, but I am guessing both are probably impossible to perceive / even see a flicker? Per your last comment, since there is no data fetching, I dont think this would help indicate the client's network loading speed would it?

apps/explorer/components/Transaction/TransactionHeader.tsx
```diff
@@ -27,14 +32,16 @@ export function TransactionHeader({ title, transactionHeaderData }: Props) {
...
         {!!timestamp && (
-          <Tooltip content={timestamp}>
-            <Text font="mono" color="subtle" ellipsis>
-              {humanDate(timestamp, {
-                dateStyle: 'medium',
-                timeStyle: 'short',
-              })}
-            </Text>
-          </Tooltip>
+          <ClientSideOnly fallback={<div className="animate-pulse">...</div>}>
```
alexfreska, line 35: I do prefer seeing a good amount of duplication before abstracting, but might be enough instances here for a <ClientDate /> to make sense.

alexfreska, line 35: Non-blocking comment, just wanted to add some thoughts:





Wouldn't skeleton block with roughly the same height and width as the resulting text, with animate pulse better match the resulting value and space?





Seeing them in the column of values I do like how they look there, I can see the appeal as long as its not causing layout shift - in this case I think the layout would not be affected by their width.

telestrial: I went ahead and played with this. It looks like the quirk here is that Skeleton's height doesn't always play in the layout the same way a text of exactly the same height would play. The good news is that you can play with that Skeleton height and get the correct CLS. All of these have been switched to Skeletons, across the app.

alexfreska (approved):

telestrial:
### Merge activity

* **Feb 21, 1:58 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/SiaFoundation/web/937).

PR #973, opened 2025-03-04
feat(explored-types): add v2 host type support

telestrial:

PR #974, opened 2025-03-04
feat(explorer): add v2 host type support

### LINKED ISSUE #971: Renterd: Host address not showing up in UI for V2 hosts
After a host announces itself as v2, the response field to use for the address is no longer `netAddress` but instead `v2SiamuxAddresses`.

telestrial:
This adds support for the v2 Host type in the explorer and closes #971.

There are two sharper edges here I see worth discussing, of the same ilk:

* Down stack, in the types update, we are going with a different typing strategy than `explored`--instead of being additive to `ExplorerHost`, we break this out into two types. One for v1 and one for v2, discriminating on `v2: boolean`.
* We also add a lot of `host.v2 ? ...` logic across the host slice of the explorer. I don't think this would be avoided by using one additive type. We'd instead be checking for `settings`, `priceTable`, and `rhpV4Settings` all the same. I've created some helpers around getting the netAddress and the settings as they pertain to checking whether they exist (is this a legacy assumption? It doesn't match the `explored` types for either one--that `settings` wouldn't exist).

One final note: the tests downstack from this PR will probably fail, and that's because the zen allow height has passed and we're seeing the underlying data change. This will probably mean we need to take this PRs tests as final truth.

### DIFF
--- a/apps/explorer/tsconfig.json
+++ b/apps/explorer/tsconfig.json
@@ -11,7 +11,10 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "incremental": true,
-    "types": ["jest", "node"],
+    "types": [
+      "jest",
+      "node"
+    ],
     "plugins": [
       {
         "name": "next"
@@ -27,7 +30,11 @@
     "next-env.d.ts",
     ".next/types/**/*.ts",
     ".next-development/types/**/*.ts",
-    ".next-development-testnet-zen/types/**/*.ts"
+    ".next-development-testnet-zen/types/**/*.ts",
+    "../../dist/apps/explorer/.next/types/**/*.ts"
   ],
-  "exclude": ["node_modules", "jest.config.ts"]
+  "exclude": [
+    "node_modules",
+    "jest.config.ts"
+  ]
 }
--- a/.changeset/brown-news-draw.md
+++ b/.changeset/brown-news-draw.md
@@ -1,0 +1,5 @@
+---
+'explorer': minor
+---
+
+Added support for v2 Hosts.
--- a/apps/explorer/app/host/[id]/opengraph-image.tsx
+++ b/apps/explorer/app/host/[id]/opengraph-image.tsx
@@ -8,6 +8,7 @@ import { truncate } from '@siafoundation/design-system'
 import { CurrencyOption, currencyOptions } from '@siafoundation/react-core'
 import { to } from '@siafoundation/request'
 import { getExplored } from '../../../lib/explored'
+import { getHostNetAddress } from '../../../lib/hostType'
 
 export const revalidate = 0
 
@@ -52,11 +53,11 @@ export default async function Image({ params }) {
     )
   }
 
-  if (!host.settings) {
+  if (host.v2 ? !host.rhpV4Settings : !host.settings) {
     return getOGImage(
       {
         id: host.publicKey,
-        title: host.netAddress,
+        title: getHostNetAddress(host),
         subtitle: truncate(host.publicKey, 30),
         initials: 'H',
         avatar: true,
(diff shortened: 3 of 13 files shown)


telestrial:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/974?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#974** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/974?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/974?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#973** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/973?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#972** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/972?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

apps/explorer/tsconfig.json
```diff
@@ -27,7 +30,11 @@
     "next-env.d.ts",
     ".next/types/**/*.ts",
     ".next-development/types/**/*.ts",
-    ".next-development-testnet-zen/types/**/*.ts"
+    ".next-development-testnet-zen/types/**/*.ts",
```
alexfreska, line 33: not sure why this was added twice, you should be able to remove the duplicate and also the last ..dist one.

apps/explorer/lib/hostTypeHelpers.ts
```diff
@@ -0,0 +1,20 @@
+import { ExplorerHost } from '@siafoundation/explored-types'
+
```
alexfreska, line 2: Could drop "helpers" from the file name as its implied.

alexfreska (approved):
- Left a minor comment.
- Re the discriminated types: make sense to me 👍 
- Agree on the v2 logic, I think it would be the same regardless of type. The one alternative would be checking once at the very top and having completely separate v1 and v2 components, but I think the way you have it is probably easier.
- Could you clarify what you mean with the netaddress legacy assumption question?
- Re downstack tests: makes sense 👍

telestrial:
> Could you clarify what you mean with the netaddress legacy assumption question?

That was actually about `settings`, and I apologize for the weird wording. We do this at least once, but maybe another place or two as well:

`host.settings ? <Component /> : <ADifferentComponent />`

Host/index, for example. There, we are saying, in the false case, "your host isn't scanned, yet. Come back later." But I suspect now what would happen is that settings would always exist but everything would be some form of 0. @chris124567: Can you verify that behavior? A host not scanned would have a settings key but the value would be an object that is all/mostly 0s?

telestrial:
### Merge activity

* **Mar 5, 12:01 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/SiaFoundation/web/974).

chris124567:
" But I suspect now what would happen is that settings would always exist but everything would be some form of 0. @chris124567: Can you verify that behavior? A host not scanned would have a settings key but the value would be an object that is all/mostly 0s? Or would we never even see that because a host not scanned wouldn't return on a hostByPubkey request?

A host that hasn't been scanned yet would have totalScans = 0 and 0 failedInteractions and 0 successfulInteractions.  It would still be returned by the hosts endpoint if you provided the pubkey because we do still have some information about the host (the net address[es]).  All the settings / prices would be zero though.

telestrial:
It may be a good course, then, to move these couple few areas where we're making decisions based on the existence of settings to instead use some combination of totalScans, successfulInteractions, and failedInteractions. Will look into what we're doing, why, and go from there. Thanks!

PR #981, opened 2025-03-10
feat(exploer): use v2Settings for host page display

telestrial:

Issue #977, opened 2025-03-04
[Explorer] beta.siascan.com unable to scan sia2.howitts.co.uk:9982


NickJH:
If I try to scan my server I get:

![Image](https://github.com/user-attachments/assets/504adef4-0967-48b2-ae7b-2576d32136f7)

It works fine on the live site.

telestrial:
Thank you!

 For clarity: it's likely your host is already scanned. What's your ed25519 pubkey? We should lcheck out what the searchbar gives us for that. I suspect it will go to the right place.

I believe the issue is that the search feature (that I rewrote) is not handling this use case--searching for a host by its netAddress. I'll look into the explored API to verify, but I believe what I'll need to do is search hosts with by URL to grab the pubkey and then navigate to the host route from there.

NickJH:
My public key is `ed25519:210ad99f0e16a02ef055fdd0fcb6506106afeaab81b6834de96d6187ebae0c42`. You can check it on the live mainnet siascan site. 

I get the same issue with my other host, hostd-mim.howitts.co.uk:9982. I cannot scan it by FQDN but it works by public key - `ed25519:944add113bca332ee36b8ac49c62ead29baaa5e8a76481a164fd7be1fe5fcbd9`.

PR #982, opened 2025-03-10
feat(explorer): add host netadress search

telestrial:


telestrial:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/982?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#982** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/982?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/982?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#981** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/981?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

apps/explorer/components/Layout/Search.tsx
```diff
@@ -57,6 +57,29 @@ export function Search() {
         return
       }
 
+      // Catch possible host address, make a different request, go there.
+      if (values.query.includes('.')) {
```
n8maninger, line 61: IPv6 addresses do not include `.`

n8maninger: Not sure we want to support this since host addresses are not guaranteed to be unique, unlike public keys. Realistically, you'd also need a results page to link to the public key page for each match.

apps/explorer/components/Layout/Search.tsx
```diff
@@ -57,6 +57,29 @@ export function Search() {
         return
       }
 
+      // Catch possible host address, make a different request, go there.
+      if (values.query.includes(':')) {
```
n8maninger, line 61: Host public keys can include `:`

apps/explorer/components/Layout/Search.tsx
telestrial: > host addresses are not guaranteed to be unique

Has this always been true? Not sure if [this](https://github.com/SiaFoundation/web/commit/1b1952e51f5d8ff5d15a255cfd071c88aacc2eec#diff-9f265412b062d2d73aaeabd0b198d6b7f17d43598f012e714e322e2862e4db73) link will work, but, before I reworked this file, we were taking the 0th element from Sia Central's response under the hosts key.

apps/explorer/components/Layout/Search.tsx
telestrial: I believe we are past prefixed host pubkeys at this points in the logic of the block, though. We check for ed25519 before this (I should probably add ':' to the end of that to guard against the unlikely but possible hash match) and then route and return; check for ":", route and return; then flow down to the search request because we then can't tell what we have and must check.

apps/explorer/components/Layout/Search.tsx
n8maninger: Yeah. This has always been true. It would be pretty much impossible for it to happen naturally, but there have been a few times where someone announced another host's netaddress, I believe, to try and confuse renters, but renters always use the host's key so it doesn't actually accomplish anything.

Screenshot 2025-03-11 at 7 46 28 AM

It's better to only support look ups by public key, imo. If a user looks up their host by address after a malicious user announces the same address, the user would be shown the malicious host instead of theirs. If the user announces their host again, it will correct itself, but that's unnecessary for the network to function.

telestrial: >  If a user looks up their host by address after a malicious user announces the same address, the user would be shown the malicious host

How would you feel if we handled this case? If I understand correctly, hostsList would return an array of Hosts > 1. Here, we could do nothing, alert something like "Multiple results found. Please use the host's public key.", and reset the form.

For context on why I wanted to add this, we have a [ticket](https://github.com/SiaFoundation/web/issues/977) from a tester that tried to search by URL. It appeared to be one of the first things they tried to do. Their host is scanned, but not supporting this created a perception for them that it wasn't.

The desire behind this feature likely turns on "I know my URL" and "how far away am I (physically/electronically) from my hostd instance?" I can open tabs, type credentials, and copy and paste with the best of them, but I can see the "race" here being won by simply typing your URL out, in a lot of cases.

If it's a convenience we can safely offer, I would like to.

n8maninger: Personally, I'm in favor of resetting expectations over implementing something that may or may not work. Having a flow that isn't rock solid for users is bad UX long-term, even if it's uncomfortable short-term imo. Having said that, I'll defer to you on implementation here.

telestrial: I certainly understand that position. I will proceed forward here with my middle ground, but this will be on my radar. I'll be periodically checking the discord/issues for anything around this. If this feature is a problem, we can always remove it. As a potential note to my future self: if we find that users searching by host ADDRESS:PORT tend to reach the "multiple hosts found" case often, we would likely need to remove this as a feature or come up with some other solution.

apps/explorer/components/Layout/Search.tsx
```diff
@@ -51,12 +51,47 @@ export function Search() {
       }
 
       // Catch possible host pubkey, avoid request, go there.
-      if (values.query.slice(0, 7) === 'ed25519') {
+      if (values.query.slice(0, 8) === 'ed25519:') {
```
peterjan, line 54: can't believe you frontenders don't have `HasPrefix`

apps/explorer/components/Layout/Search.tsx
peterjan: I'm not terribly familiar with the search in the explorer as well as the database but this is definitely a little flaky. Maybe we should consider adding a validator that checks the query against a regex? That has problems of its own of course. Alternatively we could always treat it as a net address and only go to the host page if we find a host, at the cost of an extra request for every search of course.. definitely not ideal.

apps/explorer/components/Layout/Search.tsx
telestrial: You know what. I just looked through MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith. Not sure I've ever used it by fits really well here.

peterjan (approved):

telestrial:
### Merge activity

* **Mar 21, 2:34 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/SiaFoundation/web/982).

PR #1018, opened 2025-03-27
feat(explorer): add support for v2 types

telestrial:


apps/explorer/app/contract/[id]/page.tsx
```diff
@@ -54,38 +96,75 @@ export default async function Page({ params }) {
...
+    if (!formationTxn)
+      throw new Error('No formation transaction found in successful request')
+
+    if (renewalTxnError) throw renewalTxnError
+    if (!renewalTxn)
+      throw new Error('No renewal transaction found in successful request')
+
+    formationTransaction = formationTxn
+    // renewalTransaction = renewalTxn
+    renewedFromID = formationTxn.fileContractRevisions?.[0].revision.id
```
Comment (telestrial) [inline apps/explorer/app/contract/[id]/page.tsx:165]: Here, we're typed to V2Transaction, which has `fileContractRevisions` and `fileContracts` as `V2FileContract` in `core/v2.ts`. Trying as best I can to use those core types but this disagrees.

apps/explorer/lib/contracts.ts
```diff
@@ -23,3 +30,75 @@ export function determineContractStatus({
 
   return successful ? 'obligation succeeded' : 'obligation failure'
 }
+
+export type ContractData = {
+  confirmationIndex: ChainIndex
+  fileMerkleRoot: string
+  filesize: number
+  hostBurned: (SiacoinOutput & { id: SiacoinOutputID }) | undefined
```
telestrial, line 38: We use the id on the contracts page as copyable values beneath `Renter payout: remaining renter allowance`, `Host payout: payout minus risked collate`, etc. The given type by explored doesn't give that id. But I'm not sure how important that is. This is not a clickable value that takes you anywhere. How useful is this information? Would the address be more valuable? If we didn't use added `id`, I could simply use `SiacoinOutput` here.

apps/explorer/app/contract/[id]/page.tsx
```diff
@@ -24,97 +42,58 @@ export const revalidate = 0
...
-  if (!contract) return notFound()
-
-  if (previousRevisionsError) throw previousRevisionsError
-  if (!previousRevisions)
-    throw new Error('No previousRevisions found in successful request')
-
-  if (currentTipError) throw currentTipError
-  if (!currentTip) throw new Error('No currentTip found in successful request')
+  let contract: ExplorerFileContract | ExplorerV2FileContract
+  let contractRevisions: ExplorerFileContract[] | ExplorerV2FileContract[]
```
Comment (alexfreska) [inline apps/explorer/app/contract/[id]/page.tsx:46]: I think it would be simpler to do the full v1 and v2 and return in their respective if else blocks.

apps/explorer/lib/fetchChainData.ts
```diff
@@ -0,0 +1,180 @@
...
+}
+
+export function fetchAddressSiacoinUTXOs(address: string) {
+  return fetchOrThrow(() =>
+    getExplored().addressSiacoinUTXOs({ params: { address } })
+  )
+}
+
+// Block
+export function fetchBlockByID(id: string) {
```
alexfreska, line 66: I think it makes sense to have a short hand for these cases like notFound, but what does explored actually return if you try to fetch a block ID that does not exist? Does it come back as empty data or a 400 error or a 500 error? Also many of the endpoints where you are not fetching a resource by ID like consensus tip either return data or error so if they did return empty data that would be an error rather than  404 not found.

but I dont think all of these are technically needed. "404 not found" is when the pages ID resource is not found, and other required data like consensus tip will never be empy/not found, and if it does return empty data that would be an error, eg:
```
  const [block, currentTipInfo] = await Promise.all([
    fetchBlockByID(id),
    fetchConsensusTip(),
  ])
 ```
 Maybe a more explicit wrapper could be a good extension of this:
 ```
   const [block, currentTipInfo] = await Promise.all([
    dataOrNotFound(getExplored().blockByID({ params: { id } })),
    getExplored().consensusTip(),
  ])
  ```
  Where we only wrap the resources the main resources with "404" with dataOrNotFound (or whatever the best name is), and the method checks for the correct 400 error or if explored returns empty data when not found it checks for that. The rest of the methods don't need any wrapper as they can just naturally throw an error and generate a 500 page.

File: apps/explorer/lib/contracts.ts (hunk 2 above)
Comment (alexfreska) [reply apps/explorer/lib/contracts.ts:38]: Ok how about we just remove that behaviour if its not that useful, and simplify these to just output types.

File: apps/explorer/lib/fetchChainData.ts (hunk 4 above)
Comment (telestrial) [reply apps/explorer/lib/fetchChainData.ts:66]: Thank you for this comment. It's a great one and has me thinking wide about user expectation vs. api response/design

> what does explored actually return if you try to fetch a block ID that does not exist?

It 404s with a string response.

--

After thinking through your comment and looking through this some, I think I should do the following:

1. Eliminate the notFound on empty logic and instead always throw. I don't know how I got to this, but I think that `contractRevisionByID` would still return the 0th revision, at a contract's inception. In other words, it will never be empty--meaning there are no requests where empty is acceptable--at least as far as the explorer is concerned.

2. I should drill through the fetch wrappers a boolean to fetchOrThrow that says, "This is a primary fetch". If it is, we do notFound on the 404 but then throw for other errors. If not a primary fetch, we always throw.

How does that strike you? It seems to me that we want to throw on everything that isn't a 404 on a "primary" resource so that users know a 404 means the ID is likely bad.

Comment (alexfreska) [reply apps/explorer/lib/fetchChainData.ts:66]: Yeah, thats pretty much along the lines of what I was thinking, maybe you meant this already but the only clarification I would add is that I don't think you need to add a boolean to the current method, rather you can simplify the method to solely be about checking for 400 error and calling notFound - and only wrap primary resources. No need to wrap all the other non-primary resource fetch methods with anything since they will naturally throw an error. That should give the method a very explicit and clear focus, whenever you see it used its super clear that these one or two requests are for finding the primary resources based off an external input ID (the route ID etc), these other ones are not wrapped with it because they are just supporting data that we always expect will be returned.

Comment (telestrial) [reply apps/explorer/lib/fetchChainData.ts:66]: Could we take a step back and revisit the implementation of `to()`?

Even with the changes above, minus the primary drilling thing, we would still have to do this everywhere:

> if (thingError || !thing) throw new Error('some overly broad error');

This is because the current `to` implementation returns a tuple where both `data` and `error` are optionally present — even though, for our purposes, only one will ever exist.

```ts
export async function to<T>(
  promise: Promise<AxiosResponse<T>>
): Promise<
  [
    AxiosResponse<T>['data'] | undefined,
    AxiosError<T> | undefined,
    AxiosResponse<T> | undefined
  ]
> {
  try {
    const response = await promise;
    return [response.data, undefined, response];
  } catch (error) {
    return [undefined, error as AxiosError<T>, undefined];
  }
}
```

The 0th index, in success, can’t be `undefined`, and, in error, can’t be `AxiosResponse<T>`. Similar sort of thing in the 1st index between error and undefined. It's not a bug. It's not really even a mistype. But it is an overly broad type. We'll never return `[undefined, undefined, ...]`, but, at least in the explorer, we were writing quite a few lines to handle that:

```ts
const [data, error, response] = await to(getExplored().consensusTip());
// data is typed T | undefined
if (error) throw error;
// data is still typed T | undefined
if (!data) throw new Error('No data!');
// data is now typed T
```

What would you think about this implementation of `to`?

```ts
type Success<T> = [data: T, error: undefined, response: AxiosResponse<T>];
type Failure<T> = [data: undefined, error: AxiosError<T>, response: AxiosResponse<T> | undefined];

export async function to<T>(
  promise: Promise<AxiosResponse<T>>
): Promise<Success<T> | Failure<T>> {
  try {
    const response = await promise;
    return [response.data, undefined, response];
  } catch (error) {
    const axiosError = error as AxiosError<T>;
    return [undefined, axiosError, axiosError.response]; // adding response here to check status codes
  }
}
```

Usage:

```ts
const [data, error, response] = await to(getExplored().consensusTip());
// data is typed T | undefined
if (error) throw error; // or return notFound
// data as T
```

I know that's a step above what we're talking about, but we're really talking about the consequences of the way this function returns when we could probably narrow this and simplify anything below it, wherever it is used.

apps/explorer/app/contract/[id]/page.tsx
```diff
@@ -25,104 +36,81 @@ export const revalidate = 0
...
+      fetchBlockByID(formationTxnChainIndices[0].id),
+      fetchConsensusTip(),
+    ])
+
+    return (
+      <ContractView
+        contractRevisions={contractRevisions}
+        currentHeight={currentTip.height}
+        contract={contract}
+        // renewedFromID={renewedFromID ? stripPrefix(renewedFromID) : renewedFromID}
```
Comment (peterjan) [inline apps/explorer/app/contract/[id]/page.tsx:101]: can these be removed?

apps/explorer/components/Contract/index.tsx
```diff
@@ -14,30 +14,34 @@ import { ContractHeader } from './ContractHeader'
 import {
   ChainIndex,
   ExplorerFileContract,
-  FileContractID,
-  SiacoinOutput,
+  ExplorerV2FileContract,
+  // FileContractID,
```
peterjan, line 18: can this be removed?

apps/explorer/app/tx/[id]/opengraph-image.tsx
```diff
@@ -14,65 +28,50 @@ export const size = {
...
+      subtitle: 'transaction',
+      initials: 'T',
+    },
+    size
+  )
+
 export default async function Image({ params }) {
   const id = params?.id as string
+  let searchResult: SearchResultType
+  let transaction: ExplorerTransaction | ExplorerV2Transaction
```
Comment (alexfreska) [inline apps/explorer/app/tx/[id]/opengraph-image.tsx:45]: Some more opportunities to simplify the v1 vs v2 variables and logic.

apps/explorer/app/tx/[id]/opengraph-image.tsx
```diff
@@ -14,65 +28,50 @@ export const size = {
...
-        params: {
-          id,
-        },
-      })
-    ),
-    to(getExplored().consensusTip()),
-  ])
+  // Do we have a v1 or v2 transaction?
+  try {
+    searchResult = await fetchSearchType(id)
```
Comment (alexfreska) [inline apps/explorer/app/tx/[id]/opengraph-image.tsx:50]: Can move all logic in here to simplify reassignment, can reduce to a single catch at the bottom.

apps/explorer/app/tx/[id]/page.tsx
```diff
@@ -23,30 +35,25 @@ export const revalidate = 0
...
+  const searchResult = await fetchSearchType(id)
 
-  if (transactionError) throw transactionError
-  if (transactionChainIndicesError) throw transactionChainIndicesError
-  if (currentTipError) throw currentTipError
-  if (!transaction || !transactionChainIndices || !currentTip) return notFound()
+  // Hit the v1 or v2 transaction-related endpoints.
+  if (searchResult === 'transaction') {
+    transaction = await fetchV1Transaction(id, { primaryRequest: true })
+    transactionChainIndices = await fetchV1TransactionChainIndices(id)
```
Comment (alexfreska) [inline apps/explorer/app/tx/[id]/page.tsx:47]: Ditto.

apps/explorer/lib/fetchChainData.ts
alexfreska: Yeah I like that, thats much better and there are many extra conditionals we can then remove. Lets add that in a separate PR directly to `main` (I think all existing code will still be valid), and then this stack/PR can streamline the conditionals. 

But also I think the discussion in this thread means we can remove / don't need `to` in most places.

telestrial: Will do, on getting this `to` change into a separate pr.

> But also I think the discussion in this thread means we can remove / don't need to in most places.

Oh, I didn't think about that but you may be right. If we're just going to write `if (thingError) throw throwError` where thingError is a caught error in `to`, why catch it at all? It's naturally going to hit Explorer's ErrorBoundary without our intervention. This is a great point. We then write a function that specifically handles the primary requests and checks for the 404, probably using `to`. Will try this out and see if there's an unforeseen edges but that all sounds right and great to me. Thanks!

apps/explorer/app/contract/[id]/page.tsx
Comment (telestrial) [reply apps/explorer/app/contract/[id]/page.tsx:101]: See comment on #1026

apps/explorer/components/Contract/index.tsx
telestrial: See comment in #1026

apps/explorer/lib/fetchChainData.ts
telestrial: Got this change in. Without `to`, the files seem more verbose, which I don't love, but the rest of the assumptions above were pretty solid.

peterjan (approved):

apps/explorer/app/block/[id]/page.tsx
```diff
@@ -38,30 +38,27 @@ export default async function Page({ params }) {
...
-      to(getExplored().consensusTip()),
+      getExplored().consensusTip(),
     ])
 
-  if (blockError) throw blockError
-  if (currentTipInfoError) throw currentTipInfoError
-  if (!block) return notFound()
-  if (!currentTipInfo) throw notFound()
+  if (blockError) {
+    if (blockResponse?.status === 404) return notFound()
```
Comment (alexfreska) [inline apps/explorer/app/block/[id]/page.tsx:59]: 👍

alexfreska (approved):

telestrial:
### Merge activity

* **Apr 4, 3:02 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/SiaFoundation/web/1018).

PR #1024, opened 2025-03-31
feat(explorer): sync and index warnings

alexfreska:
- The explorer now shows specific syncing and indexing warnings.

### DIFF
--- a/apps/explorer/lib/sync.ts
+++ b/apps/explorer/lib/sync.ts
@@ -7,3 +7,16 @@ export function getIsSynced(data?: { prevTimestamps: string[] }) {
         Date.now() - hoursInMilliseconds(12)
     : false
 }
+
+export function getIsIndexing(
+  consensusState?: { index: { height: number } },
+  blockMetrics?: { index: { height: number } }
+) {
+  if (!blockMetrics || !consensusState) {
+    return false
+  }
+
+  // Only show indexing warnings if the node is more than 2 blocks behind on indexing.
+  const padding = 2
+  return blockMetrics.index.height < consensusState.index.height - padding
+}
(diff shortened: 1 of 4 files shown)


alexfreska:
* **#1024** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/1024?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/1024?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

telestrial (approved):

apps/explorer/lib/sync.ts
```diff
@@ -7,3 +7,16 @@ export function getIsSynced(data?: { prevTimestamps: string[] }) {
...
+
+export function getIsIndexing(
+  consensusState?: { index: { height: number } },
+  blockMetrics?: { index: { height: number } }
+) {
+  if (!blockMetrics || !consensusState) {
+    return false
+  }
+
+  // Only show indexing warnings if the node is more than 2 blocks behind on indexing.
```
peterjan, line 19: Is 2 blocks sufficient? I guess it would be unusual but we often mine a bunch of blocks in short succession, not sure how long indexing a block takes...

alexfreska: I was thinking it would be enough to prevent the index warning from popping up every block during normal operation, and in rare cases where it does drop behind more than 2, its actually good to show the warning. I also suspect the indexing will be fast enough that this will not happen, but if that ends up wrong we can adjust this value.

PR #1033, opened 2025-04-04
Set up contracts in clusterd

chris124567:
If we have network=v1 or network=v2, we will:
- Fund a new address by mining a block
- Use it to create 3 contracts
    - One that we let expire (`expiredContract`)
    - One that we revise and then submit a storage proof for (`successfulContract`)
    - One that we renew (`renewedContract`)
- And expose these contract addresses via `/fixtures/explored`.

Example successful response:
```
{"expiredContract":"664aa3a708fce114eebb6ff352af8956c45aaaf0cb6b55f9808a3af7e5137af3","successfulContract":"297048b6a99a484598d1ce41c71b8ac05377cf71e6d03e0db7d51a596f37db04","renewedContract":"0a44d067f23bc315211439376b85297afc133cd0ac453043f86a758ffa8d5fb5"}
```
Example unsuccessful response:
```
{"expiredContract":null,"successfulContract":null,"renewedContract":null}
```

I think it is best to do it in Go vs JS because for v2 creating the contracts is already serious pain in the neck.


internal/cluster/cmd/clusterd/contract.go
```diff
@@ -0,0 +1,482 @@
...
+	}
+	for _, cau := range applied {
+		for i := range fces {
+			cau.UpdateElementProof(&fces[i].StateElement)
+		}
+	}
+
+	id := txn.ID()
+	fc1ID, fc2ID, fc3ID := txn.V2FileContractID(id, 0), txn.V2FileContractID(id, 1), txn.V2FileContractID(id, 2)
+	return FixtureContractAddresses{
```
alexfreska, line 477: Maybe have these APIs return the whole contracts since we will be fetching them anyway. And just double checking, those contract struct would have ID on them right?

alexfreska (commented):
This is great, should make testing each very straightforward! Left one comment.

alexfreska:
@chris124567 could you squash your commits and update the message to follow conventional commits, eg:

`feat(cluster): add explored contract fixtures`

internal/cluster/cmd/clusterd/contract.go
chris124567: Yeah they would

chris124567:
Added v1/v2 endpoints

`/fixtures/explored/v1`
`/fixtures/explored/v2`

which return `types.FileContractElement`s and `types.V2FileContractElement`s respectively instead of just IDs.

internal/cluster/cmd/clusterd/contract.go
```diff
@@ -0,0 +1,489 @@
...
+	sfc := (consensus.State{}).SiafundCount()
+	tm := mod64(target, sfc)
+	gm := mod64(guess, sfc)
+	if gm.Cmp(tm) < 0 {
+		guess = guess.Sub(types.NewCurrency64(sfc))
+	}
+	return guess.Add(tm).Sub(gm)
+}
+
+func mineBlocks(log *zap.Logger, cm *chain.Manager, n int) {
```
ChrisSchinnerl, line 71: Use `nm.MineBlocks` on the nodes manager `nm` instead. That automatically keeps the whole cluster in sync.

internal/cluster/cmd/clusterd/contract.go
```diff
@@ -0,0 +1,489 @@
...
+//
+//	   payout = renterPayout + hostPayout + payout*tax
+//	∴  payout = (renterPayout + hostPayout) / (1 - tax)
+//
+// This would work if 'tax' were a simple fraction, but because the tax must
+// be evenly distributed among siafund holders, 'tax' is actually a function
+// that multiplies by a fraction and then rounds down to the nearest multiple
+// of the siafund count. Thus, when inverting the function, we have to make an
+// initial guess and then fix the rounding error.
+func taxAdjustedPayout(target types.Currency) types.Currency {
```
ChrisSchinnerl, line 30: We avoid copying core logic. Use `State.FileContractTax` and `V2FileContractTax` instead.

internal/cluster/cmd/clusterd/contract.go
```diff
@@ -0,0 +1,489 @@
...
+			continue
+		} else if err := cm.AddBlocks([]types.Block{b}); err != nil {
+			log.Panic("failed to add funding block", zap.Error(err))
+		}
+		n--
+		log.Debug("mined block", zap.Stringer("index", cm.Tip()))
+	}
+}
+
+func setupV1Contracts(log *zap.Logger, cm *chain.Manager, pk types.PrivateKey, fundingBlock types.Block) FixtureV1ContractAddresses {
```
ChrisSchinnerl, line 84: Passing `fundingBlock` is a bit odd considering you have a wallet. I would just rely on the wallet to fund the contract rather than hacking together something using the miner payouts of a block directly.

It probably also makes sense to create the wallet outside of this function and then pass it to both setup functions unless you need different private keys to own these contracts.

internal/cluster/cmd/clusterd/main.go
```diff
@@ -164,9 +183,33 @@ func main() {
 	go s.Run()
 
 	nm := nodes.NewManager(dir, cm, s, nodes.WithSharedConsensus(true))
+	handler := api.Handler(cm, s, nm, log.Named("api"))
+
+	var mu sync.Mutex
+	var v1ContractAddresses FixtureV1ContractAddresses
+	var v2ContractAddresses FixtureV2ContractAddresses
+	mux := http.NewServeMux()
+	mux.HandleFunc("/fixtures/explored/v1", func(w http.ResponseWriter, r *http.Request) {
```
ChrisSchinnerl, line 192: Do we really need these routes? The explorer should be aware of existing contracts itself right? 
If I understand correctly the whole idea of adding the contract creation to the startup of the explorer was to avoid having to add routes to form contracts within tests.

internal/cluster/cmd/clusterd/wallet.go
```diff
@@ -0,0 +1,39 @@
...
+	})
+}
+
+func setupWallet(cm *chain.Manager, pk types.PrivateKey) (*wallet.SingleAddressWallet, *testutil.EphemeralWalletStore, error) {
+	ws := testutil.NewEphemeralWalletStore()
+	sw, err := wallet.NewSingleAddressWallet(pk, cm, ws)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to create renter wallet: %w", err)
+	}
+	defer sw.Close()
```
ChrisSchinnerl, line 33: closing it before even returning it?

internal/cluster/cmd/clusterd/wallet.go
```diff
@@ -0,0 +1,39 @@
...
+
+	if err := syncWallet(cm, sw, ws); err != nil {
+		return nil, nil, fmt.Errorf("failed to scan wallet: %w", err)
+	}
+	return sw, ws, nil
```
ChrisSchinnerl, line 38: I'd consider adding a little wrapper to make this a little less C-like to use. 

```go
type wallet struct {
  *wallet.SingleAddressWallet
  cm *chain.Manager
  ws *testutil.EphemeralWalletStore
}

func newWallet(cm *chain.Manager, pk types.PrivateKey) (*wallet, error) {...}
```

and then have a `Sync()` method on it

internal/cluster/cmd/clusterd/contract.go
```diff
@@ -0,0 +1,489 @@
...
+	if len(fces) != 3 {
+		log.Panic("Wrong number of contracts in block, expected 3, got", zap.Int("len", len(fces)))
+	}
+	// sort by revision number so fc[0] = fc1, fc[1] = fc2, ...
+	sort.Slice(fces, func(i, j int) bool {
+		return fces[i].V2FileContract.RevisionNumber < fces[j].V2FileContract.RevisionNumber
+	})
+	for _, cau := range applied[1:] {
+		for i := range fces {
+			cau.UpdateElementProof(&fces[i].StateElement)
```
ChrisSchinnerl, line 344: Could you avoid this manual proof updating by asking the explorer for the latest element?

ChrisSchinnerl (changes requested):

internal/cluster/cmd/clusterd/contract.go
chris124567: I use V2FileContractTax but could not get FileContractTax to work.  Not sure if there's a bug but this doesn't work:
```
    contractPayout := renterPayout.Add(hostPayout)
    contractPayout = contractPayout.Add(cm.TipState().FileContractTax(types.FileContract{Payout: contractPayout}))

    // contract we will let expire
    fc1 := types.FileContract{
        // we could use joint unlock conditions with renter and host key, but
        // this is simpler
        UnlockHash:  uc.UnlockHash(),
        WindowStart: cm.Tip().Height + 1,
        WindowEnd:   cm.Tip().Height + 2,
        Payout:      contractPayout,
        ValidProofOutputs: []types.SiacoinOutput{
            {Value: renterPayout, Address: renterAddr},
            {Value: hostPayout, Address: hostAddr},
        },
        MissedProofOutputs: []types.SiacoinOutput{
            {Value: renterPayout, Address: renterAddr},
            {Value: hostPayout, Address: hostAddr},
            {Value: types.ZeroCurrency, Address: types.VoidAddress},
        },
    }

```

```
{"error": "transaction 7433fac921d6f62e3ae4987dd4d7c17ca49b573db02f5119d5b809bf4e2b7cb9 is invalid: file contract 0 has payout with incorrect tax"}
```

internal/cluster/cmd/clusterd/main.go
chris124567: They are not strictly necessary.  We can remove the endpoints, but then in the testing code they'll have to look a couple blocks back and find the contract IDs. I guess I will remove them since Nate seems to be in favor of that as well.

internal/cluster/cmd/clusterd/contract.go
ChrisSchinnerl: Do you mine beyond the `Network.HardforkTax` before forming the contracts? Because that hardfork height should be the only thing affecting the output.

But maybe you can actually just use `rhp.PrepareContractFormation` from `core`. That way you don't even need to construct the `types.FileContract` manually.

internal/cluster/cmd/clusterd/wallet.go
```diff
@@ -0,0 +1,46 @@
...
+	"fmt"
+	"math"
+
+	"go.sia.tech/core/types"
+	"go.sia.tech/coreutils/chain"
+	"go.sia.tech/coreutils/testutil"
+	"go.sia.tech/coreutils/wallet"
+)
+
+type swallet struct {
```
peterjan, line 13: could avoid `swallet` which is a little odd imo by aliasing coreutils wallet as `cwallet` here

internal/cluster/cmd/clusterd/contract.go
```diff
@@ -0,0 +1,382 @@
...
+	proto2 "go.sia.tech/core/rhp/v2"
+	proto4 "go.sia.tech/core/rhp/v4"
+	"go.sia.tech/core/types"
+	"go.sia.tech/coreutils/chain"
+	"go.sia.tech/explored/api"
+	"go.uber.org/zap"
+)
+
+func mineBlock(nm *nodes.Manager, e *api.Client, w *swallet) error {
+	if err := nm.MineBlocks(context.Background(), 1, types.VoidAddress); err != nil {
```
peterjan, line 20: did you change your editor settings 😅 seems like you use 2 spaces for a tab now

peterjan (approved):

ChrisSchinnerl (approved):

chris124567:
It turns out that for the `transition` network, depending on how many renterd/walletd/hostd nodes we spin up, we might be before the require height or after it by the time the contract formation code runs.  We could check the current height and decide to either form v1 contracts or v2 contracts, but that seems like it could cause confusion.  So I just decided to disable contract formation on `transition`.

PR #1052, opened 2025-05-07
feat(explorer): add infinite scroll to the the address page

telestrial:


telestrial:
* **#1052** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/1052?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/1052?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

apps/explorer/app/api/explored/address-events/route.ts
```diff
@@ -0,0 +1,26 @@
...
+  const limit = searchParams.get('limit') || '20'
+  const offset = searchParams.get('offset') || '0'
+
+  if (!address) {
+    return NextResponse.json({ error: 'Address is required' }, { status: 400 })
+  }
+
+  try {
+    const { data: events } = await getExplored().addressEvents({
+      params: { address, limit: Number(limit), offset: Number(offset) },
```
peterjan, line 16: should you harden against these parameters being NaN here?

apps/explorer/components/Address/index.tsx
```diff
@@ -44,12 +42,53 @@ export function Address({
...
+        params: {
+          address: id,
+          limit: 100,
+          offset: offset,
+        },
+      })
+
+      const moreEvents = response.data
+
+      if (moreEvents?.length) {
```
peterjan, line 75: I think you could benefit from another state param called `exhausted`. You could `setExhausted` to `true` if the events you receive are fewer than the `limit` you passed. In which case you could update the UI accordingly perhaps. I don't know, feel free to resolve.

apps/explorer-e2e/src/specs/address.spec.ts
```diff
@@ -75,7 +75,7 @@ test('address displays the intended data', async ({ page }) => {
   await explorerApp.goTo('/address/' + address)
   await expect(page.getByText(`Address ${address.slice(0, 5)}`)).toBeVisible()
   await expect(page.getByText(events.data[0].id.slice(0, 5))).toBeVisible()
-  await expect(page.getByText(`${events.data.length} events`)).toBeVisible()
+  expect(await page.getByTestId('entity-link').count()).toBe(events.data.length)
```
alexfreska, line 78: iirc this will not wait for the expected count, it evaluates count() immediately. I think you are looking for:
```
await expect(page.getByTestId('entity-link')).toHaveCount(events.data.length)
```

apps/explorer/components/LoadingTimestamp.tsx
```diff
@@ -1,12 +1,17 @@
...
   children,
+  width = '200px',
 }: {
   children: React.ReactNode
+  width?: string
 }) {
   return (
-    <ClientSideOnly fallback={<Skeleton className="w-[200px] h-[20px]" />}>
+    <ClientSideOnly
+      fallback={<Skeleton className={clsx('h-[20px]', `w-[${width}]`)} />}
```
alexfreska, line 13: Whats the reason for clsx here?

alexfreska, line 13: tailwind cannot create a class dynamically like this, probably best to make the prop `className` and pass the whole `w-[130px]` so tailwind can detect it and generate the class.

apps/explorer/components/Entity/EntityList.tsx
```diff
@@ -26,10 +29,46 @@ export function EntityList({
   emptyMessage,
   skeletonCount = 10,
   children,
+  onScrollThroughMiddle,
+  customHeight,
```
alexfreska, line 33: `customHeight` => `className`

apps/explorer/components/Address/index.tsx
```diff
@@ -44,12 +42,53 @@ export function Address({
...
+  const [confirmedEvents, setConfirmedEvents] = useState<ExplorerEvent[]>([
+    ...initialEvents,
+  ])
+  const [unconfirmedEventsState] = useState<ExplorerEvent[]>([
+    ...unconfirmedEvents,
+  ])
+  const [loading, setLoading] = useState(false)
+  const [offset, setOffset] = useState(initialEvents.length)
+
+  const fetchMoreEvents = async () => {
```
alexfreska, line 60: Maybe try `useSWRInfinite` for this: https://swr.vercel.app/docs/pagination.en-US#useswrinfinite

apps/explorer/app/api/explored/address-events/route.ts
```diff
@@ -0,0 +1,26 @@
...
+export async function GET(request: Request) {
+  const { searchParams } = new URL(request.url)
+  const address = searchParams.get('address')
+  const limit = searchParams.get('limit') || '20'
+  const offset = searchParams.get('offset') || '0'
+
+  if (!address) {
+    return NextResponse.json({ error: 'Address is required' }, { status: 400 })
+  }
+
```
alexfreska, line 13: Rather than implementing a extra wrapper endpoint, use type `explored` sdk directly from the client like the search feature.

apps/explorer/components/Address/index.tsx
```diff
@@ -44,12 +42,53 @@ export function Address({
...
+    if (loading) return
+    setLoading(true)
+
+    try {
+      const response = await axios.get(`/api/explored/address-events`, {
```
alexfreska, line 65: Full comment above - use typed sdk directly.

alexfreska (commented):
Nice, this will be great, left some feedback! and I'll try it out on the preview once the explorer is syncing.

apps/explorer/app/api/explored/address-events/route.ts
telestrial: When we changed this api to the `getExplored()` syntax, we added `next/headers` to read a cookie. That's a server-side only api. I run into:

`  × You're importing a component that needs next/headers. That only works in a Server Component which is not supported in the pages/ directory. Read more: https://nextjs.org/docs/getting-started/
  │ react-essentials#server-components
  │ 
  │ 
   
╭─[.../explored.ts:1:1]
 1 │ import { Explored } from '@siafoundation/explored-js'
 2 │ import { exploredApi } from '../config'
 **3 │ import { cookies } from 'next/headers'**
`

apps/explorer/components/Address/index.tsx
telestrial: Not a bad idea. Got it!

apps/explorer/components/LoadingTimestamp.tsx
telestrial: This was supposed to be set up differently. Great catch. Thank you!

telestrial: This was supposed to be setup differently. I was prototyping and didn't clean it up. Thank you!

apps/explorer/components/Address/index.tsx
telestrial: Checked it out and got it in!

apps/explorer/components/LoadingTimestamp.tsx
```diff
@@ -1,12 +1,22 @@
...
 }: {
   children: React.ReactNode
+  width?: string
 }) {
   return (
-    <ClientSideOnly fallback={<Skeleton className="w-[200px] h-[20px]" />}>
+    <ClientSideOnly
+      fallback={
+        <Skeleton
+          className={clsx('h-[20px]', width ? `w-[${width}]` : 'w-[200px]')}
```
alexfreska, line 15: tailwind scans to codebase to understand what classes to generate and inject in the resulting css, it will find any static class names in the codebase but it will not be able to generate a class for string variable concatenation like `w-[${width}]`. It will work if you add a `className` prop and pass the full `w-[130px]`.

apps/explorer/components/Entity/EntityList.tsx
```diff
@@ -1,35 +1,77 @@
...
+  onScrollThroughMiddle,
+  className,
 }: Props) {
   const showHeading = title || actions
   let show = 'emptyState'
 
+  const listRef = useRef<HTMLDivElement>(null)
+  const debounceTimeout = useRef<NodeJS.Timeout | null>(null)
+
+  useEffect(() => {
```
alexfreska, line 43: Check out the Intersection Observer API, its a browser API that should help streamline this. There are a few places in the repo we use it with the following hook: `import { useInView } from 'react-intersection-observer'`, it has a `threshold` param that can be used to specify the percentage visible to trigger. https://github.com/thebuilder/react-intersection-observer

telestrial:
@alexfreska I've been using `a2d8b179748b8b040247c5c0205d333757dd810e76619d3256d926bde6411afe291d384de4ae`

apps/explorer/components/Entity/EntityList.tsx
telestrial: One detail that may be getting lost here is that I'd like to fetch more events when a user is scrolled 50% or more down the list (see line 48). I don't think the IO API is a good tool for that purpose. We can place a trigger div halfway through the list (which is already a bit of work to place, recalculate, and re-place as opposed to the pure event list) but, if the user scrolls past where that trigger div would be re-placed, they defeat the signal that tells the app to load more. You would reach the "bottom" of what's loaded and not know that there were more events. You would scroll up to trigger more loading. It gets messy. Maybe then you make the trigger div span the entire back half of the list, z-index'd out of the way, but, at that point of div mutation, I search for other options.

With scrollHeight, we get a much more consistent signal--even if a user drags the scrollbar down to the bottom, the moment they reach 50%, even before they up click, we're loading more events. No matter how or how fast they scroll, if >= 50% down the list => load more.

I do see the IO API being a great solution if I wanted this to trigger at the bottom, but my thought is to anticipate a bit more. Part of the illusion of infinite scroll to me is never reaching the  bottom unless you're truly at the bottom.

telestrial: In taking this out of the scrolled window, as per your other request on vercel, my previous comment is no longer relevant. Went ahead and added the div to the end and used `useInView`.

telestrial:
Went ahead and moved this to non-scrolling window and used the Intersection Observer on a div at the end.

apps/explorer/app/address/[id]/page.tsx
```diff
@@ -24,14 +24,13 @@ export const revalidate = 0
 export default async function Page({ params }) {
   const address = params?.id as string
 
+  // Fetch initial 20 events only
```
Comment (peterjan) [inline apps/explorer/app/address/[id]/page.tsx:27]: `nit` remove

peterjan (approved):
LGTM 👍

alexfreska (approved):
@telestrial nice updates, looks good!

PR #1064, opened 2025-05-17
refactor(explorer): typescript strict mode

telestrial:
This refactor for typescript strict mode was fairly straightforward. Biggest change by number of files touched was simply typing out `params: { id: string }`. I unified that prop type and made it extensible, just in case it ever needs to be. Beyond that:

* The rankHosts function needed a decent rewrite, and I suspect that, previous to this, it was at least somewhat broken. I believe we were referencing keys that didn't exist, and I'm guessing that was skewing the final result. Should I throw a changeset on that, since the resultant list is now an entirely different group of hosts?

* Satisfying strict mode for the Transaction slice was fairly verbose. Lots of novel `'key' in object ? ... : ...`. It does probably need consolidating earlier. There is some decision-making to do there about what a consolidated type should care about, due to all the custom types used (and the available keys/types therein), so I've resolved to leave that for another day.

* The `identicon.js` import in `/lib/avatar.ts` has no official typings. It does have an `@types`, but that didn't seem to be the correct typing. Could be a versioning issue, but the end result is that I `ts-expect-error`'d this import. We could alternatively declare a module in a `*.d.ts` file in libs.

### DIFF
--- a/.changeset/green-spiders-behave.md
+++ b/.changeset/green-spiders-behave.md
@@ -1,0 +1,5 @@
+---
+'explorer': minor
+---
+
+Improved top host ranking.
--- a/apps/explorer/app/address/[id]/opengraph-image.tsx
+++ b/apps/explorer/app/address/[id]/opengraph-image.tsx
@@ -2,6 +2,7 @@ import { humanSiacoin, humanSiafund } from '@siafoundation/units'
 import { getOGImage } from '../../../components/OGImageEntity'
 import { truncate } from '@siafoundation/design-system'
 import { getExplored } from '../../../lib/explored'
+import { ExplorerPageProps } from '../../../lib/pageProps'
 
 export const revalidate = 0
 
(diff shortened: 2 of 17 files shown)


telestrial:
* **#1064** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/1064?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/1064?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

alexfreska (approved):
Good stuff! A few comments on the items you outlined:
- rankHosts: if the algorithm changed a bit, yeah add a changeset saying it was improved so results may be slightly different now.
- Transaction verbose: I think I would lean towards forking out v1 or v2 very early and separate components, that way never worry about v1 handling again and can build all future features with purely v2 data - but normalizing is probably quite viable as well.
- identicon.js typing: I think adding a .d.ts would be the conventional solve here, give it a try otherwise all good.

PR #1065, opened 2025-05-20
feat(clusterd): add networkVersion parameter to startCluster

telestrial:

PR #1066, opened 2025-05-20
test(explorer): cover all network versions

telestrial:


telestrial:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/1066?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1066** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/1066?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/1066?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1065** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/1065?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

apps/explorer-e2e/src/specs/contract.spec.ts
```diff
@@ -6,55 +6,128 @@ import {
...
+        await explorerApp.navigateBySearchBar(contractID)
+
+        await expect(
+          page
+            .getByTestId('entity-heading')
+            .getByText('Contract ' + contractID.slice(0, 5))
+        ).toBeVisible()
+      }
+
+      if (networkVersion === 'v2') {
```
peterjan, line 66: Instead of duplicating the test implementation for every network version, I'd write one implementation of the test and then use a switch on the network version where it's appropriate or needed. That way we don't needlessly duplicate a bunch of code but it also serves as documentation, highlighting the differences between the network versions.

peterjan (commented):
is there a lot of value in running all tests for every version regardless whether the test itself is affected by it? I didn't expect to see a change to the block spec for instance, or the home spec. Really only matters in the contract spec I think, might be missing something though.

apps/explorer-e2e/src/specs/contract.spec.ts
alexfreska: @telestrial similar to the discussions in the code for contracts/transactions, I think it would be most clear to just split out separate tests rather than weaving conditionals. It doesn't duplicate that much and will make it easier to leave the v1 test alone and continue to extend the v2 features and test in the future with no concern over v1.

apps/explorer-e2e/src/specs/address.spec.ts
```diff
@@ -14,90 +14,102 @@ import { exploredStabilization } from '../helpers/exploredStabilization'
 let explorerApp: ExplorerApp
 let cluster: Cluster
 
-test.beforeEach(async ({ page, context }) => {
-  cluster = await startCluster({ context })
-  await renterdWaitForContracts({
-    renterdNode: cluster.daemons.renterds[0].node,
-    hostdCount: cluster.daemons.hostds.length,
-  })
+const networkVersions = ['v1', 'v2', 'transition'] as const
```
alexfreska, line 17: We can drop `transition`, it won't be any different from testing against v1 or v2 unless the test explicitly it written to mine through the network transition.

alexfreska (commented):
Re @peterjan's comment: I like the goal of exhaustive v1 and v2 coverage right now just to smoke test everything works as we make final changes and move through the hardfork. But duplicating all the tests does add significant CI time, so maybe we create a ticket to drop the more redundant v1 tests down the line, and just leave the ones with an explicit difference, contract and tx.

telestrial:
In terms of v1 and v2: there are differences across all routes. Address probably the least so, but we do expect to see and display v2 contract and transaction events there. Block, home, and host use different keys out of the same routes. Contract and transaction are entirely different types out of different routes.

I think we use different code paths for each slice of this, to some extent.

telestrial:
### Merge activity

* **May 30, 4:36 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/SiaFoundation/web/1066).

PR #1075, opened 2025-05-27
Update alerts data

n8maninger:
Adds the current and latest versions from the versioning alert to the display.


telestrial (changes requested):
I suspect we need to make a change to the value render here, and then, if you want to pass the test-js test, you could reword the commit to something like this:

`feat(hostd): add versioning data to the alerts table`

telestrial (approved):

PR #1100, opened 2025-06-13
feat(explorer): add output route, display, and search

telestrial:
Fairly straightforward. Tests included. There were some minor cascading visual tweaks to the outputs list on the transaction page. Since both the ID and address are now clickable (previously only the output title or icon linked to the address), I added title beside these, for clarity. Those extra characters put some strain on the responsiveness of the ID/hash flex container, so I brought down the number of characters displayed in those copyables and adjusted the layout to stack the ID and address more aggressively at smaller viewport widths, rather than hiding the address entirely.

### DIFF
--- a/.changeset/eighty-pumas-explode.md
+++ b/.changeset/eighty-pumas-explode.md
@@ -1,0 +1,5 @@
+---
+'explorer': minor
+---
+
+Added support for output search and display.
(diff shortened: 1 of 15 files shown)


libs/hostd-types/src/api.ts
```diff
@@ -248,6 +248,8 @@ export type V2ContractsPayload = {
 }
 
 export type V2ContractsResponse = {
+  id(id: any): unknown
```
alexfreska, line 251: This doesn't look related.

telestrial: No idea how that happened. Great catch!

alexfreska (approved):

telestrial:
### Merge activity

* **Jun 24, 1:09 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/SiaFoundation/web/1100).

PR #1101, opened 2025-06-14
Update SitePageHead.tsx

timrettop:
This should resolve an issue requesting the site manifest returning a 401 behind a reverse proxy.  https://medium.com/@aurelien.delogu/401-error-on-a-webmanifest-file-cb9e3678b9f3

### DIFF
--- a/apps/website/components/SitePageHead.tsx
+++ b/apps/website/components/SitePageHead.tsx
@@ -47,7 +47,7 @@ export function SitePageHead({
           sizes="16x16"
           href="/favicon-16x16.png"
         />
-        <link rel="manifest" href="/site.webmanifest" />
+        <link rel="manifest" href="/site.webmanifest" crossorigin="use-credentials" />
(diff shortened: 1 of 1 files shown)


alexfreska:
@timrettop thank you for the PR! How did you notice this as an issue? I would like to try to recreate it.

timrettop:
This would require fronting the service with a reverse proxy like Traefik or nginx proxy. 
Then open developer tools on the browser and connect to the HTTPS Proxy URL. 
You should see an error 401 unauthorized when the client fetches the manifest URL.

alexfreska:
@timrettop does the new site https://sia.tech (now in a separate repo that is not yet public) give you this issue?

timrettop:
Hmm, looking at the html for ```https://sia.tech/``` I don't see a /site.webmanifest defined.

alexfreska:
@timrettop ah you are right I can add it. Also wanted to clarify just in case, did you also hit this issue with renterd or hostd?

timrettop:
Hostd

PR #1105, opened 2025-06-18
feat(explorer): add maturity height indicator to address page events

telestrial:
This one was a bit tricky in terms of placement. I ended up re-writing the flex containers in EntityListItem (before, two rows in flex-col. now, one flex-row row with flex-col containers). Beyond that flex strategy getting rewritten, I decided to place this at the end but with a small border to the left. It disappears at the `sm` breakpoint.

### DIFF
--- a/apps/explorer/components/Address/index.tsx
+++ b/apps/explorer/components/Address/index.tsx
@@ -31,6 +31,7 @@ type Tab = 'events' | 'utxos'
 
 type Props = {
   id: string
+  networkHeight: number
   addressInfo: {
     balance: AddressBalance
     unconfirmedEvents: ExplorerEvent[]
@@ -40,6 +41,7 @@ type Props = {
 
 export function Address({
   id,
+  networkHeight,
   addressInfo: {
     balance: { unspentSiacoins, unspentSiafunds },
     unconfirmedEvents,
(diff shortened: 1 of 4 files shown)


apps/explorer/components/Address/index.tsx
```diff
@@ -279,45 +308,55 @@ function formatPayoutEntity(payout: ExplorerEvent): EntityListItemProps {
...
+  siacoinOutput: ExplorerSiacoinOutput,
+  currentHeight: number
 ): EntityListItemProps {
   return {
     hash: siacoinOutput.id,
     sc: new BigNumber(siacoinOutput.siacoinOutput.value),
     label: 'siacoin output',
     initials: 'SO',
     scVariant: 'value' as const,
+    maturityData: {
```
alexfreska, line 328: Two little tweaks:
- we should add an `isMature` here that is `const isMature = e.maturityHeight <= nodeBlockHeight` and use that to render locked or unlocked.
- For the bottom height in the UI, instead of current network height we should pass the height of the specific event
That way the bottom number is the height when the txn event occurred, the top number is when it unlocks, and the lock vs unlock is determined by the network height.

alexfreska (approved):
Nice, looks good!

telestrial:
### Merge activity

* **Jun 24, 1:09 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/SiaFoundation/web/1105).

PR #1104, opened 2025-06-17
fix: sidenav logo firefox flex and alignment

alexfreska:
- Sidenav Logo is now properly positioned on Firefox browsers.

Screenshot 2025-06-17 at 3 41 01 PM

PR #1108, opened 2025-06-18
fix(renterd): improve uploads experience

alexfreska:
- The app is now more stable and responsive when uploading files especially on localhost or http/1.1 and Firefox-based browsers.

## Issues
- Users were experiencing very wonky app behaviour during uploads, the main issue was that multipart uploads were saturating the browsers limit of 6 requests in http/1.1 and this meant other requests such as navigating pages or between directories were being blocked from going through. With bad host contracts this could block further requests almost indefinitely as renterd kept reattempted to upload stalled parts.
- In addition the multipart slot logic appeared to somehow have had slightly different behaviour on Firefox, that was causing additional erroneous requests.

## Fix
- The app now limits the number of concurrent multipart requests based on http version, with a max of 5 on http/1.1.
- The multipart uploader has been refactored with a much clearer loop which has removed the Firefox edge cases.
- The multipart uploader slot/concurrency system is now global spanning all multipart uploads. Previously each upload had its own slots so one upload stalled out on its final part would block all its slots.
- Refactored and added to the tests.

### renterd with lousy contracts
```
2025-06-19T10:43:11-04:00       DEBUG   worker.worker.uploadmanager     sector upload failure was penalised     {"uploadError": "sector upload already finished; failed to upload sector to contract 9eb01e5d51e3411897be0816f33549be086508f77a13f072c571e30c078142bc; failed to write sector: failed to calculate root: stream was gracefully closed", "uploadDuration": "0s", "totalDuration": "8.547291458s", "overdrive": false, "penalty": 85470, "hk": "ed25519:de13e12e37bb3662caeb3a78e7159281aa5c8cf1f160c88792cea393cd4410fc"}
2025-06-19T10:43:17-04:00       DEBUG   worker.worker.uploadmanager     sector upload failure was penalised     {"uploadError": "sector upload already finished; failed to upload sector to contract e617ad43a1d9e121221d820ac5a68d18b0c6ab59b1926e69aec49020e7f956e9; failed to write sector: failed to read response: stream was gracefully closed", "uploadDuration": "0s", "totalDuration": "37.592916625s", "overdrive": false, "penalty": 375920, "hk": "ed25519:91bf3019f428f3affe45dcb10b904e5f4f3f33da368d9316d32b4fe0e4bec223"}
```

### DIFF
--- a/apps/renterd/lib/multipartUpload.ts
+++ b/apps/renterd/lib/multipartUpload.ts
@@ -28,7 +28,6 @@ export type MultipartParams = {
     busUploadAbort: ApiBusUploadAbort
   }
   partSize?: number
-  maxConcurrentParts?: number
   onProgress?: (event: {
     sent: number
     total: number
@@ -44,12 +43,51 @@ type UploadedPart = {
 }
 
 export class MultipartUpload {
+  // Static properties for global control
+  static #globalMaxConcurrentParts = 5
+  static #activeSlots = 0
+  static #slotQueue: Array<() => void> = []
+
+  static get activeSlots(): number {
+    return MultipartUpload.#activeSlots
+  }
+
+  static setGlobalMaxConcurrentParts(max: number): void {
+    if (max < 1) {
+      throw new Error('globalMaxConcurrentParts must be at least 1.')
+    }
+    MultipartUpload.#globalMaxConcurrentParts = max
+  }
+
+  private static async acquireSlot(): Promise<void> {
+    if (
+      MultipartUpload.activeSlots < MultipartUpload.#globalMaxConcurrentParts
+    ) {
+      MultipartUpload.#activeSlots++
+      return
+    }
+
+    // Wait for a slot to become available
+    return new Promise<void>((resolve) => {
+      MultipartUpload.#slotQueue.push(resolve)
+    })
+  }
+
+  private static releaseSlot(): void {
+    MultipartUpload.#activeSlots = Math.max(MultipartUpload.#activeSlots - 1, 0)
+
+    const nextUpload = MultipartUpload.#slotQueue.shift()
+    if (nextUpload) {
+      MultipartUpload.#activeSlots++
+      nextUpload()
+    }
+  }
+
   // params
   #bucket: string
   #key: string
   #file: File
   #partSize: number
-  #maxConcurrentParts: number
   #api: {
     workerUploadPart: ApiWorkerUploadPart
     busUploadComplete: ApiBusUploadComplete
@@ -72,6 +110,7 @@ export class MultipartUpload {
   #uploadedParts: UploadedPart[]
   #uploadId?: string
   #aborted: boolean
+  #completed: boolean
   // error retry backoff
   #initialDelay = 500 // 1/2 second
   #maxDelay = 60_000 // 1 minute
@@ -82,7 +121,6 @@ export class MultipartUpload {
     this.#bucket = options.bucket
     this.#key = options.key
     this.#partSize = options.partSize || 1024 * 1024 * 5
-    this.#maxConcurrentParts = Math.min(options.maxConcurrentParts || 5, 15)
     this.#file = options.file
     this.#api = options.api
     this.#onProgress = options.onProgress || (() => null)
@@ -96,6 +134,7 @@ export class MultipartUpload {
     this.#uploadedParts = []
     this.#uploadId = undefined
     this.#aborted = false
+    this.#completed = false
   }
 
   public async create() {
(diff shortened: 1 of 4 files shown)


alexfreska:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/1108?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1108** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/1108?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/1108?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1107** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/1107?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1106** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/1106?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1104** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/1104?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

apps/renterd/lib/multipartUpload.ts
```diff
@@ -325,7 +393,7 @@ export class MultipartUpload {
 
       const uploadedPart = {
         partNumber: partNumber,
-        // removing the " enclosing characters from the raw ETag
+        // Remove the " enclosing characters from the raw ETag.
         eTag: eTag.replace(/"/g, ''),
       }
```
telestrial, line 399: <p class='graphite__hidden'><i>[Re: line +402]</i></p>
Could there be memory benefits to deleting the part's progressCache entry in a line below here? Is it useful to have that entry, after we've succeeded and entered this finally block?

```
delete this.#progressCache[partNumber]
delete this.#activeConnections[partNumber]
```
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/1108?utm_source=unchanged-line-comment">Graphite</a>.</p>

telestrial, line 331: <p class='graphite__hidden'><i>[Re: line -334]</i></p>
Could there be memory benefits to deleting the part's progressCache entry in a line below here? Is it useful to keep this entry, after we've entered this finally block?

```
delete this.#activeConnections[partNumber]
delete this.#progressCache[partNumber]
```
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/1108?utm_source=unchanged-line-comment">Graphite</a>.</p>

apps/renterd/lib/multipartUpload.ts
```diff
@@ -44,12 +43,51 @@ type UploadedPart = {
 }
```
telestrial, line 44: <p class='graphite__hidden'><i>[Re: line -41]</i></p>
Could this exist as an array (tuple-like) instead? like `[parNumber, eTag]` ? And then you reference what you need where you need it, assuming the positionings. That is maybe a little less safe but it's only you that will use it. I would be interested about the performance gains of that vs. an object, even though the object is more convenient to address.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/1108?utm_source=unchanged-line-comment">Graphite</a>.</p>

telestrial (approved):
This looks great. I took a full review pass on `multipartUpload`, because I wanted to understand it better, and left a few thoughts. Not sure how relevant they are but I know this is important logic and the "very large file" case is particularly challenging. There may be some opportunities to slim this down, from a memory perspective, which could come in useful there.

apps/renterd/lib/multipartUpload.ts
```diff
@@ -96,6 +134,7 @@ export class MultipartUpload {
     this.#uploadedParts = []
     this.#uploadId = undefined
     this.#aborted = false
+    this.#completed = false
   }
 
   public async create() {
```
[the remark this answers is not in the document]
alexfreska: Yeah thats better, ill flip it around.

apps/renterd/lib/multipartUpload.ts
alexfreska: You are right that it would use less memory to store it as arrays, but the current structure is what the "complete" API requires, so if we changed this we would end up storing it as arrays, transforming it and allocating all the objects anyway.
But in terms of bottlenecks when dealing with a very large upload, this is definitely the right angle to evaluate, this specific list of data is required by the API so it has to scale with the number of parts, but there are probably other large lists that are currently kept in memory that could be pruned down to only tracking in-progress parts.
Beyond that the main most critical bit to keep an eye on is making sure the slices of the actual data are not kept referenced otherwise that means that all uploads are kept in memory, eg 10gb file uses 10gb, and this is also where browser engines can be different if we are not super careful, Firefox may drop the reference and garbage collect slightly different, etc.

alexfreska:
### Merge activity

* **Jun 20, 4:48 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/SiaFoundation/web/1108).

PR #1199, opened 2025-07-18
ci: add nx cache, run e2e on main

alexfreska:
> CI performance improvements, fixes, and configuration.

- Pinning daemon versions in e2e tests until they have been fixed in cluster
- Added nx cache step, this should significantly speed up many builds where large libraries like the design-system are unchanged.
  - Notes: Removed the next cache key as it should not be combined with npm or bun. Not adding a next cache for now, if we do add one it should be a cache and key per project. For now testing how only an nx cache performs, also important to note that `affected` logic means unchanged apps are already not rebuilt on PR, but they may now load from cache on many steps that build all projects such as on main workflows.
- e2e shards now run on pushes to main, previously the `if` logic meant e2e shards were skipped.
- Removed Force build SDK steps, targets that require the sdk now include `"dependsOn": ["sdk:build"]` which has the same effect.


alexfreska:
* **#1199** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/1199?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/1199?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

chris124567 (approved):

alexfreska:
Only the 2 bulk integrity check tests are failing and we cannot upgrade cluster daemons at the moment, merging.

PR #1231, opened 2025-08-01
feat(design-system): data table

alexfreska:
- Added TanStack-based DataTable and supporting hooks.
  - Initial pass at a TanStack table for use in indexd and eventually it can replace our homegrown table API in all apps.
  - Subject to change as we build out indexd and figure out what we need.
  - Currently configured to support virtualization, and client-side pagination, faceting, filtering, sorting by default.
  - Table data parameters are URL driven and uniquely scoped.
  - Table goes for more density than those in our existing apps, also aims to support displaying and manipulating thousands of rows.
  - Reason for introducing this is that it will provide solid foundation with strong ecosystem of plugins for column filtering, faceting, ordering, etc and especially virtualization - users have asked for many of these features and they will be easier to implement and maintain with this library. Also interestingly the table's API is pretty much the same as our homegrown ones.
  
### screenshot from next PR which uses it in indexd
Screenshot 2025-08-01 at 12 06 37 PM

### DIFF
--- a/.changeset/metal-guests-find.md
+++ b/.changeset/metal-guests-find.md
@@ -1,0 +1,5 @@
+---
+'@siafoundation/design-system': minor
+---
+
+Added TanStack-based DataTable.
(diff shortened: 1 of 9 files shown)


alexfreska:
### Merge activity

* **Aug 4, 10:09 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/SiaFoundation/web/1231).

PR #1247, opened 2025-08-07
fix(types): host settings protocol version type change

alexfreska:
- Updated the protocol version types.
  - V2HostSettings and indexd config both now have a v prefix in protocol version fields, eg: `v1.1.1`.
  - Double checking if the renterd autopilot config minProtocolVersion is affected.


telestrial (changes requested):
See vercel comment. I'm scratching my head a bit because your change should work? Is it possible there's some caching issue and we're still using the tuple somehow?

PR #1257, opened 2025-08-11
refactor: change contract height labels for greater clarity

telestrial:


telestrial:
* **#1257** <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/1257?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/SiaFoundation/web/1257?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

apps/explorer/components/Contract/index.tsx
```diff
@@ -76,22 +89,32 @@ export function Contract({
...
+            label: 'resolution type',
+            value:
+              contractResolutionLabels[
+                contract.resolutionType as keyof typeof contractResolutionLabels
+              ] ?? '-',
+            copyable: false,
+          },
       {
-        label: 'proof confirmed',
+        label: 'Resolution Txn',
```
alexfreska, line 106: Looks like we have a mix of lower case, capital case, sentence case on the labels

apps/explorer/components/Contract/index.tsx
```diff
@@ -48,6 +48,19 @@ export function Contract({
   })
   const values = useMemo(() => {
     return [
+      'hostPublicKey' in contract && {
+        label: 'Host public key',
+        entityType: 'hostPublicKey',
+        entityValue: contract.hostPublicKey,
+        displayValue:
+          contract.hostPublicKey?.replace('ed25519:', '').slice(-12) + '...',
```
alexfreska, line 56: Are we sure we want to remove the ed25519 prefix?

telestrial: I'm not sure, no. It made more visual sense when i was looking at adding the renter public key here as well. You've got 12 preview characters and so that prefix would be eight of them. The link and clipboard copy are still correct here. I'm not married to removing them but that was my thinking.

alexfreska: My instinct is that since its a part of the value probably a little confusing to remove it.

PR #1292, opened 2025-08-29
feat(explorer): add troubleshooting tool

telestrial:
Adds a troubleshooter tool to the explorer, using the Sia Central API.

### DIFF
--- a/apps/explorer-e2e/src/specs/troubleshooter.spec.ts
+++ b/apps/explorer-e2e/src/specs/troubleshooter.spec.ts
@@ -1,0 +1,37 @@
+import { test, expect } from '@playwright/test'
+import { ExplorerApp } from '../fixtures/ExplorerApp'
+import {
+  foundationHostKey,
+  TROUBLESHOOTER_FORM_SUBMIT,
+  TROUBLESHOOTER_RESULTS,
+} from '../fixtures/constants'
+
+let explorerApp: ExplorerApp
+
+test.beforeEach(async ({ page }) => {
+  explorerApp = new ExplorerApp(page)
+})
+
+test('can search a host', async ({ page }) => {
+  // This pattern exists because the troubleshooter API is rate limited. We can't
+  // do multiple tests that hit this route in quick succession with the same key.
+  const requestPromise = page.waitForRequest(
+    `**/troubleshoot/${foundationHostKey}*`,
+  )
+  await page.goto('/troubleshoot')
+
+  await page
+    .getByRole('textbox', { name: 'ed25519:707d18...' })
+    .fill(foundationHostKey)
+  await page.getByTestId(TROUBLESHOOTER_FORM_SUBMIT).click()
+
+  const request = await requestPromise
+
+  expect(request.url()).toContain(`/troubleshoot/${foundationHostKey}`)
+})
+
+test('can directly navigate to a host', async ({ page }) => {
+  await explorerApp.goTo('/troubleshoot/' + foundationHostKey)
+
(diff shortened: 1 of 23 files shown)


apps/explorer/components/TroubleshooterForm/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const router = useRouter()
+  const explored = useExplored()
+
+  const form = useForm({
+    defaultValues: defaultValues,
+  })
+
+  const onSubmit = useCallback(
+    async (values: { query: string }) => {
+      setIsFetching(true)
```
alexfreska, line 56: The form has this built in, its called `isSubmitting` https://react-hook-form.com/docs/useform/formstate

apps/explorer/lib/troubleshooter.ts
```diff
@@ -0,0 +1,70 @@
...
+  dialTime: number
+  errors: string[]
+  warnings: string[]
+  handshake: boolean
+  handshakeTime: number
+  netAddress: NetAddress
+  resolvedAddresses: string[] | null
+  scanTime: number
+  scanned: true
+  settings: {
```
alexfreska, line 25: Looks like you can use `V2HostSettings` type here?

apps/explorer/lib/troubleshooter.ts
```diff
@@ -0,0 +1,70 @@
+import {
+  Address,
+  Currency,
+  NetAddress,
+  Protocol,
+} from '@siafoundation/explored-types'
+import axios from 'axios'
+
+export type TroubleShooterRequest = {
```
alexfreska, line 9: Most code uses Troubleshooter, some places such as here use TroubleShooter.

libs/design-system/src/core/Button.tsx
```diff
@@ -127,6 +127,17 @@ export const buttonStyles = cva(
           'hover:text-gray-1000 hover:dark:text-graydark-1000',
           'text-gray-700 dark:text-graydark-800',
         ],
+        amber2: [
```
alexfreska, line 130: Lets keep the action button colors semantic, ie, primary green, secondary gray, danger destructive red, etc.

I see where this comes from: because we match the primary accent green with the Sia network green, it makes buttons like this feel a little inconsistent on zen. But if we want to tackle this, I think the full fix here is to either:
- De-emphasize the relationship between current network and button colors by changing the primary accent color. I think I prefer this.
- Investigate and implement a dynamic button accent color pattern that changes per network and still makes sense with the semantic colors.
In both cases this would apply to all our apps.

apps/explorer-e2e/src/specs/troubleshooter.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+  await page.goto('/troubleshoot')
+
+  await page
+    .getByRole('textbox', { name: 'ed25519:707d18...' })
+    .fill(foundationHostKey)
+  await page.getByTestId(TROUBLESHOOTER_FORM_SUBMIT).click()
+
+  const request = await requestPromise
+
+  expect(request.url()).toContain(`/troubleshoot/${foundationHostKey}`)
```
alexfreska, line 30: Does this have a different effect then checking the resulting page URL?

telestrial:
### Merge activity

* **Sep 5, 8:33 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/SiaFoundation/web/1292).

PR #1301, opened 2025-09-03
fix(libs): estimated block height uses dynamic genesis

alexfreska:
- Estimated block height now uses dynamic genesis timestamp. https://github.com/SiaFoundation/walletd/issues/238

### DIFF
--- a/libs/renterd-react/src/bus.ts
+++ b/libs/renterd-react/src/bus.ts
@@ -11,10 +11,7 @@ import {
   delay,
   throttle,
 } from '@siafoundation/react-core'
-import {
-  getMainnetBlockHeight,
-  getTestnetZenBlockHeight,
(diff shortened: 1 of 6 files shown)


libs/renterd-react/src/bus.ts
```diff
@@ -299,12 +296,7 @@ export function useEstimatedNetworkBlockHeight(): number {
   })
   const res = useSWR(
     state,
-    () => {
-      if (state.data?.network === 'zen') {
-        return getTestnetZenBlockHeight()
-      }
-      return getMainnetBlockHeight()
-    },
+    () => getBlockHeightFromGenesis(state.data?.hardforkOak.genesisTimestamp),
```
alexfreska, line 299: This data is not actually available in renterd.

PR #1513, opened 2026-06-09
feat(indexd): add bulk delete for quotas, keys, and accounts

telestrial:

PR #1514, opened 2026-06-09
fix(indexd): clear row selection after bulk actions complete

telestrial:


telestrial:
* **#1514** <a href="https://app.graphite.com/github/pr/SiaFoundation/web/1514?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/web/1514?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1513** <a href="https://app.graphite.com/github/pr/SiaFoundation/web/1513?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

apps/indexd/dialogs/KeyQuotaReassignDialog.tsx
```diff
@@ -92,11 +94,12 @@ export function KeyQuotaReassignDialog({
             await mutate((key) => key.startsWith(adminConnectKeysRoute))
             form.reset()
             closeDialog()
+            onSuccess?.()
```
alexfreska, line 97: It would be simpler to call table.setRowSelection({}) here and the other few places inline - wouldn't need all the other plumbing.

alexfreska (approved):

apps/indexd/dialogs/KeyQuotaReassignDialog.tsx
telestrial: In the dialog case, I'm not so sure of that. Maybe it's something I don't see. You have to pass something to the dialog. You could pass table, but I think that's more or less the same writing. `onSuccess={() => table.setRowSelection({})` -> `onSuccess()` vs. `table={table}` -> `table.setRowSelection({})`. To me, the former is more clear at the call site--no clicking into the dialog to figure out what it does.  
  
There is a broader thing I investigated here, too: a fix starting in `useDataTable`, but `selectedRows`only recomputes when `rowSelection` changes. You run into problems solving the problem there because `manualPagination`means we then invalidate all the pages. It's a tricky balance here, I think, unless there's something I don't see.

telestrial:
### Merge activity

* **Jun 10, 5:01 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/SiaFoundation/web/1514).
* **Jun 10, 5:03 PM UTC**: [Graphite](https://app.graphite.com/github/pr/SiaFoundation/web/1514) rebased this pull request as part of a merge.
* **Jun 10, 5:05 PM UTC**: @telestrial merged this pull request with [Graphite](https://app.graphite.com/github/pr/SiaFoundation/web/1514).

apps/indexd/dialogs/KeyQuotaReassignDialog.tsx
alexfreska: You don't need to drill it thru all the layers, you could get table from a hook inside the dialog here, same way you get the quotas and keyUpdate etc.

telestrial: Oh duh. I may throw a fix here. Thank you.

PR #1538, opened 2026-08-19
fix(explorer): use selected currency for exchange rate fallback

telestrial:
I think this is just an oversight on the "fallback" language. A user preferring EUR would see the USD exchange rate number next to the pound icon until it resolved. I believe this cleans that up but would appreciate a good eye towards whether I'm categorizing the issue correctly.

### DIFF
--- a/apps/explorer/lib/fallback.ts
+++ b/apps/explorer/lib/fallback.ts
@@ -10,7 +10,7 @@ import { getExplored } from './explored'
 export async function buildFallbackDataExchangeRate(currency: CurrencyID) {
   const explored = await getExplored()
   const { data: rate } = await explored.exchangeRate({
-    params: { currency: 'usd' },
+    params: { currency },
   })
   return {
     // Hooks build with react-core have keys of the form:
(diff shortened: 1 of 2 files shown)


telestrial:
* **#1538** <a href="https://app.graphite.com/github/pr/SiaFoundation/web/1538?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/web/1538?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

apps/explorer/lib/fallback.ts
```diff
@@ -10,7 +10,7 @@ import { getExplored } from './explored'
 export async function buildFallbackDataExchangeRate(currency: CurrencyID) {
   const explored = await getExplored()
   const { data: rate } = await explored.exchangeRate({
```
alexfreska, line 12: Maybe add some e2e tests, one for the default case, and then one that passes a cookie and checks the the server rendered page (JS disabled) lines up - that way we know for sure.

alexfreska (approved):
If its currently showing the wrong value this fix definitely makes sense. Could use a test just to be sure.

PR #1539, opened 2026-09-11
feat(hostd): replace per-price pin toggles with a single switch

telestrial:
This PR changes hostd's configuration page to support either pinning all pinnable fields or none of them, while still respecting the legacy mixed configuration. After this PR, it will not be possible through the UI to set a **new** mixed configuration. I think it's still possible to set a mixed config through the API, though?   
  
Single field pinning toggles are gone. If you're in that legacy state, the overall pin toggle is set to the off position. I wrestled with that and can see arguments for having it in the on position. I'm open to whatever we think is best.

I tested this on my own host without saving anything and used clusterd beyond that, so more manual testing on a real hostd daemon would be useful before we release this into the wild.


telestrial:
* **#1539** <a href="https://app.graphite.com/github/pr/SiaFoundation/web/1539?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/web/1539?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ChrisSchinnerl (approved):
Look at the deployment with Zeus as the backend and it looks great. I'll leave it up to Alex to judge code quality.

alexfreska (commented):
Hey, this looks great. Keeping the mixed config instead of silently resolving it is the right call, and the notice makes it clear what's going on.

I think it could be simplified a fair bit by loading a mixed config as `shouldPinPrices: null` instead of `false`. The switch still shows off, but the form itself then knows the host hasn't picked yet, so most of the legacy tracking can go.

It would also close one gap. Right now a host with a mixed config can't save to all unpinned from the switch alone. Flipping it on and back off leaves 0 changes and Save disabled, so they have to edit some unrelated field before they can save. With `null` as the starting value, switching to off counts as a change on its own.

Left some inline notes on where, plus a couple of ideas for making the notice push hosts to resolve it.

apps/hostd/contexts/config/transform.ts
```diff
@@ -263,18 +303,19 @@ export function transformDown({
       ? {
           pinnedCurrency: settingsPinned.currency,
           pinnedThreshold: new BigNumber(settingsPinned.threshold).times(100),
-          shouldPinMaxCollateral: settingsPinned.maxCollateral.pinned,
+          // Pinning only reads as enabled when every price is pinned. A mixed
+          // configuration reads as off and is surfaced separately so the host
+          // can see which of their prices are still pinned.
+          shouldPinPrices:
```
alexfreska: What if this returns `null` for a mixed config? All pinned is `true`, none pinned is `false`, anything else is `null`. The switch renders `null` as off, so it looks the same, but the form then knows the host hasn't made a choice.

apps/hostd/contexts/config/transform.ts
```diff
@@ -133,34 +136,71 @@ export function getCalculatedValues() {
...
-  existingValues: any,
-): HostSettingsPinned {
+  existingValues: any
+  legacyPinnedPrices: LegacyPinnedPrices
+  pinningTouched: boolean
+}): HostSettingsPinned {
+  // A host with a mixed pinning configuration keeps it until they explicitly
+  // change pinning, so that unrelated edits elsewhere in the config do not
+  // silently pin or unpin their prices.
+  const preserveLegacy = !pinningTouched && isMixedPinning(legacyPinnedPrices)
```
alexfreska: With `null`, I think this whole block turns into: if `values.shouldPinPrices === null`, keep the pinned flags from `existingValues`, otherwise set all four to the switch value. Then `legacyPinnedPrices` and `pinningTouched` can come off the signature, and the call in `useOnValid` goes back to how it is on main.

apps/hostd/contexts/config/index.tsx
```diff
@@ -56,6 +58,24 @@ export function useConfigMain() {
...
+    if (!checkIfAllResourcesLoaded(resources)) {
+      return noLegacyPinnedPrices
+    }
+    return getLegacyPinnedPrices(resources.settingsPinned.data)
+  }, [resources])
+  const hasMixedPinning = useMemo(
+    () => isMixedPinning(legacyPinnedPrices),
+    [legacyPinnedPrices],
+  )
+  const { pinningTouched } = usePinningTouched({ form })
```
alexfreska: I think `usePinningTouched` can go with the `null` approach, since `null` vs `true`/`false` already tells us whether the host has made a choice. Reset puts it back to `null` too, so the notice comes back without extra work.

apps/hostd/components/Config/index.tsx
```diff
@@ -1,41 +1,43 @@
...
   })
   const pinnedCurrency = useWatch({
     control: form.control,
     name: 'pinnedCurrency',
   })
+  // While a mixed configuration is unresolved each price keeps the pinned
+  // state the daemon reports for it, so the host can see what is pinned.
+  // Without the explorer there is no exchange rate and pinned settings are
+  // never saved, so every price is configured in siacoin.
+  const isPinned = (price: keyof typeof legacyPinnedPrices) => {
```
alexfreska: With `null` this could be `pinningEnabled && (shouldPinPrices ?? daemonFlag)`, reading the flag straight from the pinned settings. That would let the `legacyPinnedPrices` memo, `noLegacyPinnedPrices` and `isMixedPinning` go. Could also fold the repeated `pinningEnabled && showMixedPinningNotice` below into one value.

apps/hostd/components/Config/index.tsx
```diff
@@ -59,155 +67,52 @@ export function Config() {
           fields={fields}
           name="pinnedThreshold"
         />
-        <PanelMenuSetting
+        {pinningEnabled && showMixedPinningNotice && (
+          <Alert>
```
alexfreska: Could we add an amber warning icon to this notice, like `Warning16` with `text-amber-600`? A mixed config is a state we want hosts to get out of, and right now the plain gray box reads more like general info.

apps/hostd/components/Config/PricingSetting.tsx
```diff
@@ -0,0 +1,78 @@
...
+          ) : (
+            <ConfigurationSiacoin name={name} form={form} fields={fields} />
+          )}
+        </div>
+      }
+    />
+  )
+}
+
+export function MixedPinningNotice() {
```
alexfreska: What about "Pin all" and "Unpin all" buttons in here that set `shouldPinPrices` to `true` or `false`? Right now getting to unpinned from a mix means flipping the switch on and then off. With buttons, either direction is one click, and together with the `null` change either one counts as a change right away. The copy on line 74 would need updating to point at the buttons.

apps/hostd/contexts/config/fields.tsx
```diff
@@ -205,9 +205,17 @@ export function getFields({
...
     },
 
-    shouldPinStoragePrice: {
-      title: 'Pin storage price',
-      description: '',
+    shouldPinPrices: {
+      title: 'Pin prices',
+      description: (
+        <>
+          Pin your prices to fixed fiat values. The daemon tracks the exchange
```
alexfreska: Could we tighten this copy a bit and make it less technical? Something like:

> Pin your prices to fixed fiat values. hostd tracks the exchange rate and keeps the siacoin price in sync with the fiat value, so your fiat revenue stays consistent as the price of siacoin moves. Pinning applies to all of your prices at once - storage, egress, ingress, and max collateral are pinned together or not at all.

Mainly saying hostd instead of the daemon, and "fiat revenue" instead of "real world revenue".

apps/hostd-e2e/src/specs/configLegacyPinning.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+  // The toggle already reads off, so turn it on and back off to make the
+  // choice to not pin explicit.
+  await setSwitchByLabel(page, 'shouldPinPrices', true)
+  await setSwitchByLabel(page, 'shouldPinPrices', false)
+
+  await expect(
+    page.getByText('Some of your prices are pinned to fiat values'),
+  ).toBeHidden()
+
+  await fillTextInputByName(page, 'storagePrice', '10')
```
alexfreska: This storage price edit is only needed because flipping on and back off leaves 0 changes. With `null` it could be dropped, and the test could check that Save is enabled right after the flip.

apps/hostd/contexts/config/types.ts
```diff
@@ -31,16 +31,41 @@ export const dnsProviderOptions: { value: DNSProvider; label: string }[] = [
...
-  shouldPinEgressPrice: false,
   egressPricePinned: new BigNumber(0),
-  shouldPinIngressPrice: false,
   ingressPricePinned: new BigNumber(0),
-  shouldPinMaxCollateral: false,
   maxCollateralPinned: new BigNumber(0),
 }
 
+/**
+ * The set of individually pinned prices as reported by the daemon. Only
```
alexfreska: Small one: this says it's only populated for a mixed config, but `getLegacyPinnedPrices` returns the daemon's flags every time, including all true. Moot if this type goes away.

telestrial:
Went ahead and implemented the `null` approach. Thank you, @alexfreska! A good amount of reduced complexity in that suggestion.

alexfreska (approved):
Changes look great. One small thing: a few files aren't Prettier formatted (`components/Config/index.tsx`, `contexts/config/transform.ts`, `contexts/config/transform.spec.ts`), so maybe run `npx prettier --write` on those before merging. Otherwise good to go.

ChrisSchinnerl (approved):

